PR #2, opened 2016-10-28
Added initial version of Princess interface.

ptrbman:
-PrincessSolver.scala
 contains the wrapper code of an (incomplete) api towards princess.
-princess.jar & scala-actors-2.11.0.jar
 A (special) version of Princess with support for Natural numbers
 (i.e., an ADT with Succ, isZero, Pred and constant zero.

Also changed SolverFactory s.t. it allows --solvers=princess to run
with the new Princess backend.


ptrbman:
Here is the initial version of the PrincessSolver. I haven't added my hack for opening files so you can't test it from commandline as-is. I also haven't done anything with the license-situation. I also didn't add any of the benchmarks I tried, but one good one is attached:
[prop_01.smt2.txt](https://github.com/epfl-lara/inox/files/558262/prop_01.smt2.txt)

(It is super simple, solvable by z3 and princess quickly and by enum slightly slower).

Cheers,
Peter

samarion:
Merged, thanks!

PR #10, opened 2017-01-17
disable assertions by default

pruemmer:
Disabling (debugging) assertions makes a significant performance difference


samarion:
LGTM.

Is there some way to get some debug output from Princess? It would be nice if we could have a `PrincessDebugSection` that could enable that output as well as the assertions from command-line.
Can we pass some logger to the princess solver spawner?

I'll merge this in the meantime.

pruemmer:
At the moment, you can enable output of all assertions to a text file (in either SMT-LIB or Scala format), using options dumpSMT and dumpScala of the SimpleAPI constructor. This does not fully support ADTs yet though

Issue #11, opened 2017-01-17
cons instead of list in the tutorial?


MikaelMayer:
Hello,
Is this an issue? In the [tutorial](https://github.com/epfl-lara/inox/blob/master/src/doc/tutorial.md), you say for the size funciton

```scala
if_ (ls.isInstOf(T(list)(tp))) {
  ... /* `1 + size(ls.tail)` */
} else_ {
  ... /* 0 */
}
```
But instead of list, shouldn't it be cons?
Especially in the remaining, there is no more reference to the consConstructor or cons anymore.

```scala
/* The `E(BigInt(i))` calls correspond to `IntegerLiteral(i)` ASTs. */
if_ (ls.isInstOf(T(list)(tp))) {
  /* The recursive call to `size` written `E(size)(tp)(...)` corresponds to
   * the AST `FunctionInvocation(size, Seq(tp), Seq(...))`.
   * Note that we refer to the symbol `tail` of the `consConstructor`'s second
   * field when building the selector AST. */
  E(BigInt(1)) + E(size)(tp)(ls.asInstOf(T(list)(tp)).getField(tail))
} else_ {
  E(BigInt(0))
}
```

If this is perfectly valid, can you explain me why and maybe I will update the tutorial to reflect that.

samarion:
Ah that should indeed be a `cons` instead of `list` (both times, also in the `asInstOf` call).

PR #12, opened 2017-01-17
Fix a few typos in the tutorial, and integrate `tut` in the build process

romac:
I tried to follow the tutorial by pasting the code snippets into a `sbt console` and found what appear to be a few typos along the way.

With this changes, the code compiles but this line
```scala
solver.assertCnstr(Not(prop))
```
fails with the following error: https://gist.github.com/romac/60bbc628fad06d1266c8a681f9b80043


samarion:
The error is possibly related to https://github.com/epfl-lara/inox/issues/11.
Could you maybe integrate that change and try again?

romac:
Sure!

romac:
I'm still getting [the same error](https://gist.github.com/romac/60bbc628fad06d1266c8a681f9b80043). 

Might that be caused by the fact that Inox cannot find Z3's native interface? (I'm on OS X, and haven't yet built ScalaZ3 by hand, but I can try doing that, and copying the .jar in the `unmanaged` directory and see if it solves anything?)

romac:
Also, would you be interested in having the tutorial/doc automatically typechecked + interpreted when building? I already have a branch that implements this with [tut](https://github.com/tpolecat/tut), and am happy to open another PR once this one gets merged.

samarion:
The solver is correctly falling back to the binary z3 in your example, so it's not linked to the native bindings. It seems the trees we generate don't type-check for some reason...

Oh, well we're trying to assert `Not(prop)` where `prop` has integer type... Duh!
We need to change `prop` to say it's >= 0:
```scala
val prop = (if_ (ls.isInstOf(T(cons)(tp))) {
  E(BigInt(1)) + E(size)(tp)(ls.asInstOf(T(cons)(tp)).getField(tail))
} else_ {
  E(BigInt(0))
}) >= E(BigInt(0))
```

samarion:
Typechecking and interpreting would be great, yes please!

romac:
Done.

```scala
scala> solver.assertCnstr(Not(prop))

scala> solver.check(SolverResponses.Simple) /* Should return `Unsat` */
res3: inox.solvers.SolverResponses.Simple.Response[solver.Model,solver.Assumptions] = Unsat
```

:)

romac:
Should I add the commit for the `tut` integration to this PR or open a new one once this one gets merged?

By the way, what style do you prefer: [with REPL output](https://gist.github.com/romac/e44e01d35eac119f8099c36dcd25ba09) or [keep it the same](https://gist.github.com/romac/3d0782db09a38b46e27d56bbcba6a1aa)?

samarion:
As you wish for this or an other PR.
I think in general I'd prefer to keep it the same, adding the REPL output seems a bit too verbose in most cases.

romac:
Alright, here it is! `tutQuick` will now be run automatically after `compile` and put the rendered documentation in `doc/`.

romac:
Note: For some reason `tut` doesn't like

```scala
implicit val symbols = NoSymbols
  .withFunctions(Seq(sizeFunction))
  .withADTs(Seq(listSort, consConstructor, nilConstructor))
```

so I had to change it to:

````scala
implicit val symbols = {
  NoSymbols
    .withFunctions(Seq(sizeFunction))
    .withADTs(Seq(listSort, consConstructor, nilConstructor))
}
```

src/main/doc/tutorial.md
```diff
@@ -0,0 +1,206 @@
...
+val prop = (if_ (ls.isInstOf(T(cons)(tp))) {
+  E(BigInt(1)) + E(size)(tp)(ls.asInstOf(T(cons)(tp)).getField(tail))
+} else_ {
+  E(BigInt(0))
+}) >= E(BigInt(0))
+```
+__Note__: Inox will assume the inductive invariant on the recursive call to `size(xs)`.
+
+In order to verify the property, we get an instance of an Inox solver (see
+[Programs](/src/doc/API.md#programs) and [Solvers](/src/doc/API.md#solvers) for more details):
```
samarion: Shouldn't these point to /doc instead of /src/doc now?

romac: Fixed in e0eb71f6eda732ab27486c5d0e3adfabb9300fc1

samarion:
Merged, thanks!

PR #14, opened 2017-02-13
Added import explanation in tutorial

MikaelMayer:


samarion:
Since Romain setup tut for the documentation, the "real" documentation lives in /src/main/doc/ now.
Could you please move your changes there?

Also, for links within the same project, I like to avoid the "https://github.com/..." part. You can start the links with / which will point directly to the root of the github project.

samarion:
We need to write this (the tut part) down somewhere for documentation contributors. Where do you think it should go? Somewhere in the README?

MikaelMayer:
I updated the links as requested. However, I just saw that /src/main/doc/tutorial.md is different than /doc/tutorial.md in the sense that many scala code blocks were replaced by tut:silent code blocks, resulting in no more syntax highlighting. Is this wanted @romac ?
Because I just updated the links in the README.md file to point to them but now it's lesser quality.

samarion:
Haha sorry I wasn't clear: the doc *sources* are in /src/main/doc but the actual documentation is in /doc. So updates to the documentation should take place in /src/main/doc but then documentation links should point to stuff in /doc. Sorry!

MikaelMayer:
ok I reverted the changes to README then

PR #18, opened 2017-03-29
Bug fix: `case class Node(children: List[Node])`

MikaelMayer:
The previous hasInstance function was buggy and cannot cover the following case:

    case class Node(children: List[Node])

because it originally unwrapped `List[Node]` to `Set(List[Node], Node)`
and thus was blocked because Node was re-used.


samarion:
It seems like there is indeed a bug but the proposed fix is too "relaxed". Consider for example the adt
```scala
case class Node(children: (Node, Node))
```
This adt has no instance but would be accepted if we only look at the adt-typed fields. We probably need a specialized traversal for the type that skips adt type parameters.

MikaelMayer:
In my case I don't have this problem because I encode my tuples myself using custom adts for them.
Let me try a fix for that.

PR #21, opened 2017-04-07
Updated scalaz3 library for windows

MikaelMayer:
That's it


MikaelMayer:
How do you proceed already for telling to unix system to use the right binaries? because I can include them in the jar probably, but I don't know if it would work.

MikaelMayer:
@samarion I managed to fix scalaZ3 and push the appropriate library into inox/unmanaged.

samarion:
Awesome, thanks!

PR #26, opened 2017-05-08
Improvements for BitVectors

mantognini:
This is still slightly work in progress but input is welcome on a few points (see FIXMEs and TODOs). I'll add a review to guide the discussion a bit.


src/main/scala/inox/ast/TypeOps.scala
```diff
@@ -174,6 +174,8 @@ trait TypeOps {
         TupleType(ts).unveilUntyped
       }
 
+    // TODO BVType, maybe?
```
mantognini: Would computing the `typeBound` on `BVType` be useful?

src/main/scala/inox/ast/TypeOps.scala
```diff
@@ -306,6 +308,7 @@ trait TypeOps {
       case Untyped => Some(0)
       case BooleanType => Some(2)
       case UnitType => Some(1)
+      // TODO BVType => 2^size
```
mantognini: Would this fact be useful/harmful/useless  for verification?

src/main/scala/inox/solvers/smtlib/SMTLIBTarget.scala
```diff
@@ -428,6 +428,14 @@ trait SMTLIBTarget extends SMTLIBParser with Interruptible with ADTManagers {
       case BVAShiftRight(a, b)       => FixedSizeBitVectors.AShiftRight(toSMT(a), toSMT(b))
       case BVLShiftRight(a, b)       => FixedSizeBitVectors.LShiftRight(toSMT(a), toSMT(b))
 
+      case c @ BVWideningCast(e, _)  =>
+        val Some((from, to)) = c.cast // FIXME can we assume it's well typed here?
```
mantognini: Here and below, can I assume the expressions are well typed? If not, how should I report it?

src/main/scala/inox/solvers/smtlib/SMTLIBTarget.scala
```diff
@@ -503,6 +511,16 @@ trait SMTLIBTarget extends SMTLIBParser with Interruptible with ADTManagers {
...
+/*
+ *      TODO
+ *       case (FixedSizeBitVectors.BitVectorConstant(n, b), Some(BVType(size))) if b == size =>
+ *         BVLiteral(n, size)
+ *
+ *       case (SHexadecimal(h), Some(BVType(size))) if size <= 32 =>
+ *         BVLiteral(BigInt(h.toInt), size)
+ *       // FIXME SHexadecimal has no conversion for integer larger than 32 bits.
+ */
+
```
mantognini: Not sure if this is useful or not. It was added into Leon but maybe no longer required for inox/stainless?

src/main/scala/inox/solvers/z3/Z3Native.scala
```diff
@@ -249,7 +250,9 @@ trait Z3Native extends ADTManagers with Interruptible { self: AbstractSolver =>
       case Implies(l, r) => z3.mkImplies(rec(l), rec(r))
       case Not(Equals(l, r)) => z3.mkDistinct(rec(l), rec(r))
       case Not(e) => z3.mkNot(rec(e))
-      case IntLiteral(v) => z3.mkInt(v, typeToSort(Int32Type))
+      case bv @ BVLiteral(_, size) if size <= 32 => z3.mkInt(bv.toBigInt.toInt, typeToSort(BVType(size)))
+      // FIXME currently ScalaZ3 doesn't support integer larger than Int(32).
```
mantognini: This is currently a latent issue with ScalaZ3 so I'm not sure it's easily fixable... It's probably not high priority either.

src/main/scala/inox/solvers/z3/Z3Native.scala
```diff
@@ -529,17 +543,19 @@ trait Z3Native extends ADTManagers with Interruptible { self: AbstractSolver =>
           if(ts.length > 4 && ts.substring(0, 2) == "bv" && ts.substring(ts.length - 4) == "[32]") {
             val integer = ts.substring(2, ts.length - 4)
             tpe match {
-              case Int32Type => IntLiteral(integer.toLong.toInt)
+              case Int8Type  => Int8Literal(integer.toLong.toByte) // FIXME is this right? above reads "ends with [32]"...
```
mantognini: Not sure about this piece of code: `[32]` seems to refer to 32-bit so maybe it just doesn't make sense to have other integer types handled here?

src/main/scala/inox/solvers/z3/Z3Native.scala
```diff
@@ -731,8 +747,16 @@ trait Z3Native extends ADTManagers with Interruptible { self: AbstractSolver =>
         case BooleanType =>
           model.evalAs[Boolean](z3ID).map(BooleanLiteral)
 
+        /* FIXME missing value for parameter converter:
+         *       (z3.scala.Z3Model, z3.scala.Z3AST) => Option[Byte]
+         * case Int8Type =>
+         *   model.evalAs[Byte](z3ID).map(Int8Literal(_)).orElse {
+         *     model.eval(z3ID).flatMap(t => ex.get(t, Int8Type))
+         *   }
+         */
```
mantognini: Not sure how to handle this...

src/main/scala/inox/solvers/z3/Z3Unrolling.scala
```diff
@@ -96,7 +96,13 @@ trait Z3Unrolling extends AbstractUnrollingSolver { self =>
...
       val res = tpe match {
         case t.BooleanType => model.evalAs[Boolean](elem).map(t.BooleanLiteral)
-        case t.Int32Type => model.evalAs[Int](elem).map(t.IntLiteral(_)).orElse {
+        /* FIXME missing implicit value for parameter converter:
+         *       (z3.scala.Z3Model, z3.scala.Z3AST) => Option[Byte]
+         * case t.Int8Type => model.evalAs[Byte](elem).map(t.Int8Literal(_)).orElse {
+         *   model.eval(elem).flatMap(term => ex.get(term, t.Int32Type))
+         * }
+         */
+        case t.Int32Type => model.evalAs[Int](elem).map(t.Int32Literal(_)).orElse {
```
mantognini: This part seems to be a duplicate of Z3Native. Maybe some refactoring is needed/wanted?

src/test/scala/inox/evaluators/EvaluatorSuite.scala
```diff
@@ -118,20 +161,59 @@ class EvaluatorSuite extends FunSuite {
...
+
+    def bvl(x: BigInt) = BVLiteral(x, 13)
+    eval(e, GreaterEquals(bvl(7), bvl(4)))  === BooleanLiteral(true)
+    eval(e, GreaterThan(bvl(7), bvl(7)))    === BooleanLiteral(false)
+    eval(e, LessEquals(bvl(7), bvl(7)))     === BooleanLiteral(true)
+    eval(e, LessThan(bvl(4), bvl(7)))       === BooleanLiteral(true)
+  }
+
+  test("BitVector Division and Remainder") {
+    // FIXME is it intentional that Modulo are not tested?
```
mantognini: Was it intentional? Note that the test was initially called *Int Modulo and Remainder*.

src/test/scala/inox/solvers/SemanticsSuite.scala
```diff
@@ -54,8 +54,17 @@ class SemanticsSuite extends FunSuite {
...
+    check(s, Int8Literal(58),        Int8Literal(58))
+    check(s, Int32Literal(0),        Int32Literal(0))
+    check(s, Int32Literal(42),       Int32Literal(42))
+    // FIXME currently BV 8/16/32 are supported by BitvectorEncoder.
+    // check(s, BVLiteral(0, 13),       BVLiteral(0, 13))
+    // check(s, BVLiteral(58, 64),      BVLiteral(58, 64))
+    // FIXME currently BV 8/32 are supported by Z3Native.
+    // check(s, BVLiteral(0, 16),       BVLiteral(0, 16))
+    // check(s, BVLiteral(58, 16),      BVLiteral(58, 16))
+    // NOTE When/if this is fixed, add relevant tests in the sections below.
```
mantognini: This is probably not at the top of our priority, but do note the current limitations.

src/test/scala/inox/solvers/SemanticsSuite.scala
```diff
@@ -151,20 +193,49 @@ class SemanticsSuite extends FunSuite {
...
 
-    check(s, Division(IntLiteral(1), IntLiteral(-3)),   IntLiteral(0))
-    check(s, Remainder(IntLiteral(1), IntLiteral(-3)),  IntLiteral(1))
+    check(s, LessThan(Int32Literal(7), Int32Literal(4)),      BooleanLiteral(false))
+    check(s, LessThan(Int32Literal(7), Int32Literal(7)),      BooleanLiteral(false))
+    check(s, LessThan(Int32Literal(4), Int32Literal(7)),      BooleanLiteral(true))
+  }
+
+  test("BitVector Division and Remainder", filterSolvers(_, princess = true)) { ctx =>
+    // FIXME is it intentional that Modulo are not tested?
```
mantognini: idem

src/main/scala/inox/ast/TypeOps.scala
samarion: Hmm, I'm not sure. I think I'd prefer them to be disjoint in Inox. In Stainless we can maybe add auto-casting.

src/main/scala/inox/ast/TypeOps.scala
samarion: It could be useful (for small size) but I don't think it'll make much difference.

src/main/scala/inox/solvers/smtlib/SMTLIBTarget.scala
samarion: Yes, you can assume they're well typed. Inox should have complained earlier otherwise.

src/main/scala/inox/solvers/smtlib/SMTLIBTarget.scala
samarion: I don't know, it probably won't hurt to have it (and support conversions for hex that are larger than 32 bits).

src/main/scala/inox/solvers/z3/Z3Native.scala
samarion: Can't you use mkNumeral?

src/main/scala/inox/solvers/z3/Z3Native.scala
samarion: I don't know, I've never tried any tests with bv8, so I don't know what z3 will give us back... However, since the char type went down to 16 bits, this part should probably change anyway. You can make it work for arbitrary "[\d+]" ending or something.

src/main/scala/inox/solvers/z3/Z3Native.scala
samarion: I don't think it's needed, the default case will catch this.

src/main/scala/inox/solvers/z3/Z3Unrolling.scala
samarion: You can refactor if you want, but it's under my toleration limit for duplication :)

src/main/scala/inox/ast/TypeOps.scala
mantognini: Alright, I'll leave this out then.

src/main/scala/inox/ast/TypeOps.scala
mantognini: Ok, for now I leave this as is and we can look into it later if we feel it's relevant.

src/main/scala/inox/solvers/z3/Z3Native.scala
mantognini: I didn't know that one, will look into it! Thanks.

src/main/scala/inox/solvers/z3/Z3Native.scala
mantognini: Alright, will do.

src/main/scala/inox/solvers/z3/Z3Native.scala
```diff
@@ -501,50 +514,28 @@ trait Z3Native extends ADTManagers with Interruptible { self: AbstractSolver =>
       val kind = z3.getASTKind(t)
       kind match {
         case Z3NumeralIntAST(Some(v)) =>
-          val leading = t.toString.substring(0, 2 min t.toString.length)
```
mantognini: After toying a bit with this to fix some bugs, it can be noticeably simplified to this patch.

src/main/scala/inox/solvers/z3/Z3Native.scala
```diff
@@ -727,17 +718,25 @@ trait Z3Native extends ADTManagers with Interruptible { self: AbstractSolver =>
...
-          model.evalAs[Int](z3ID).map(i => IntegerLiteral(BigInt(i)))
+         /*
+          * FIXME this seems to work, but why not rely on the default evaluation instead of the specialised evaluation
+          *       for Int(32) which fails for big integers.
+          *
+          * case IntegerType =>
+          *  model.evalAs[Int](z3ID).map(IntegerLiteral(_)).orElse {
+          *    model.eval(z3ID).flatMap(ex.get(_, IntegerType))
+          *  }
+          */
```
mantognini: Feedback on this would be welcome: the removed code is buggy, the one commented out works but I'm not sure we want to use that trick (i.e. it'd be better to remove it altogether).

src/main/scala/inox/solvers/z3/Z3Unrolling.scala
```diff
@@ -92,14 +92,23 @@ trait Z3Unrolling extends AbstractUnrollingSolver { self =>
...
           model.eval(elem).flatMap(term => ex.get(term, t.Int32Type))
         }
-        case t.IntegerType => model.evalAs[Int](elem).map(t.IntegerLiteral(_))
+
+        /* FIXME this seems to work, but why not rely on the default evaluation instead of the specialised evaluation
+         *       for Int(32) which fails for big integers.
+         * case t.IntegerType => model.evalAs[Int](elem).map(t.IntegerLiteral(_)).orElse {
+         *   model.eval(elem).flatMap(ex.get(_, t.IntegerType))
+         * }
+         */
```
mantognini: This is the same patch as above.

src/test/scala/inox/evaluators/EvaluatorSuite.scala
samarion: Not sure, if you can get the tests to work with modulo that would be great :)

src/main/scala/inox/solvers/smtlib/SMTLIBParser.scala
```diff
@@ -81,6 +81,10 @@ trait SMTLIBParser {
       case _ => IntegerLiteral(n)
     }
 
+    // FIXME smtlib.theories.BitVectorConstant.unapply is broken! This is the correct implementation (i.e. no toInt)
+    case QualifiedIdentifier(SMTIdentifier(SSymbol(cst), Seq(SNumeral(size))), None) if cst startsWith "bv" =>
+      BVLiteral(BigInt(cst drop 2), size.toInt)
```
samarion: You should probably open a PR at scala-smtlib.

src/main/scala/inox/solvers/z3/Z3Native.scala
samarion: Hmm, I guess that `ex.get(_, IntegerType)` should have similar performance to `evalAs` but I'm not really sure. Why do you say it would be better to just remove it?

src/main/scala/inox/solvers/smtlib/SMTLIBParser.scala
mantognini: Yes, indeed. I'll do that next week.

src/main/scala/inox/solvers/z3/Z3Native.scala
mantognini: The "problem" is that `model.evalAs[Int](z3ID)` will fail when the `Int` cannot represent the `BigInt` and therefore rely (often) on `model.eval(z3ID).flatMap(ex.get(_, IntegerType))` (the default case) anyway. So, unless `evalAs[Int]` is significantly better for the valid values, it's quite useless.

samarion: If extraction goes through `eval` in most cases, then there is no point for this indeed. This is a bit surprising though as it's rare to have z3 report integer counterexamples outside of the int32 range...

mantognini: I guess in the end this needs to be measured... :-/ I'll leave a comment for that, and one could look into this in the future if we have time.

src/main/scala/inox/solvers/smtlib/SMTLIBParser.scala
mantognini: Done: regb/scala-smtlib#27
Once merged, I'll update inox to refer to the updated version of scala-smtlib and remove this patch.

src/test/scala/inox/evaluators/EvaluatorSuite.scala
mantognini: Tests added for Modulo on BV (and the implementation for solvers).

src/main/scala/inox/tip/TipDebugger.scala
```diff
@@ -47,9 +47,11 @@ trait TipDebugger extends Solver {
     }
   }
 
-  abstract override def assertCnstr(expr: Expr): Unit = {
+  abstract override def assertCnstr(expr: Expr): Unit = try {
     debugOut.foreach { o => o.printScript(encoder.encode(expr)) }
     super.assertCnstr(expr)
+  } catch {
+    case te: program.symbols.TypeErrorException => ctx.reporter.fatalError(te.pos, te.getMessage)
```
mantognini: Not sure it's the best place to add this catch, but it seemed to be the closest to the frontier between Inox and Stainless and at the same time required the less changes. This at least makes stainless reports errors more nicely. I noticed however that not all `Solver`s available through the `SolverFactory` were mixed in with `TipDebugger` so this is probably not the most robust solution. @samarion, maybe you have a better idea in mind?

---

Before:
```
[  Info  ]  - Now considering 'postcondition' VC for fun @?:?...
<stack trace of uncaught exception from Main>
```

Now:
```
[  Info  ]  - Now considering 'postcondition' VC for fun @?:?...
[ Fatal  ] Type error: <a proper error message>
<graceful exit>
```

src/main/scala/inox/ast/ExprOps.scala
```diff
@@ -141,8 +141,8 @@ trait ExprOps extends GenTreeOps {
       case StringConcat(StringLiteral(""), b) => b
       case StringConcat(b, StringLiteral("")) => b
       case StringConcat(StringLiteral(a), StringLiteral(b)) => StringLiteral(a + b)
-      case StringLength(StringLiteral(a)) => IntLiteral(a.length)
-      case SubString(StringLiteral(a), IntLiteral(start), IntLiteral(end)) =>
+      case StringLength(StringLiteral(a)) => Int32Literal(a.length)
+      case SubString(StringLiteral(a), Int32Literal(start), Int32Literal(end)) =>
```
samarion: This was actually a but in Inox: `StringLength` and `SubString` are defined for Integers and not bitvectors. Could you please fix this in this PR? Thanks!

mantognini: This was fixed.

src/main/scala/inox/solvers/smtlib/SMTLIBParser.scala
mantognini: I'm waiting for scala-smtlib to be available on sonatype to push an update for this.

src/main/scala/inox/tip/TipDebugger.scala
mantognini: In the end, I've removed this controversial bit: it's better to let the caller deal with these kind of errors and assume programs typecheck when being solved.

mantognini:
This PR now also offers some improvement w.r.t. positions: less of them are lost.

mantognini:
Now also updated to scala-smtlib 0.2.2 (bypassing sonatype for now, referring directly to its git repository).

samarion:
Merged, thanks!

PR #28, opened 2017-06-14
Performance improvement in Extractor

mantognini:
When sampling stainless' performance, the `deconstruct` function is listed as one of the most -- if not *the* most -- time consuming part of the overall execution. In order to optimise it, we can re-order the pattern matching cases to put the most frequent at the top.

The order proposed here is based (for the significantly most frequent ones) on [statistics](https://gist.github.com/mantognini/959bfbc05b3e6c4b49d4a4241d57ce61) acquired when verifying all stainless benchmark under `verification/valid`. On my workstation, I could see the following effect when executing `sbt it:test`:

version  | wall clock | user time
--|--------------|--------------
master | 39:30 min | 2771 s 
patched v1 | 32:52 min | 2349 s
patched v2 | 30:49 min | 2261 s

So the results are not negligible! :-)

You can also compare the results on laraquad:
 - [before: 18 min 44 sec](http://laraquad4.epfl.ch:9000/epfl-lara/stainless/build/1789) and 
 - [after: ~16 min 58 sec~ 16 min 26 sec](http://laraquad4.epfl.ch:9000/epfl-lara/stainless/build/1792)

### DIFF
--- a/src/main/scala/inox/ast/Extractors.scala
+++ b/src/main/scala/inox/ast/Extractors.scala
@@ -3,6 +3,8 @@
 package inox
 package ast
 
+import scala.collection.immutable.HashMap
+
 /** Meat and bones of tree deconstruction/reconstruction.
   *
   * Implementations provide methods to extract all useful subtrees from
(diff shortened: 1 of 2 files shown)


MikaelMayer:
Awesome !!

MikaelMayer:
Can you either merge the pull request or push directly on the main branch ?

mantognini:
I think I can. Always good to have positive peer review. :-)

However, I don't think I can publish inox to sonatype... and therefore not apply this patch to stainless without @samarion. So the question is, should I merge PRs (once I've got positive feedback from someone) anyway?

MikaelMayer:
You can still push to github even if you cannot push to sonatype.

Le mercredi 14 juin 2017, Marco Antognini <notifications@github.com> a
écrit :

samarion:
I'll be reading and reviewing all PRs, so I can publish to sonatype when
necessary.

About this PR, thanks this is cool. We can even do it better by using a map
from class to case. I started to write a macro about this at some point but
it's hard to be general. You can write a specialized version for inox tree
case matches which should be easy. See the macros in abide for an example.

Otherwise, see with Manos how he optimized the recursive evaluator in Leon.

mantognini:
I'll keep in mind the idea of using a Map. It indeed [looks](https://github.com/epfl-lara/leon/blob/topic/grammars/src/main/scala/leon/evaluators/TableEvaluator.scala) as a good alternative. I'll try to hack something quickly with that. But I'll stay away of the macro for now as there are more important optimisation/refactoring to be made I think.

> I'll be reading and reviewing all PRs

Good to know, I'll let you merge things then! :-)

mantognini:
BTW, I've updated the description with the new runtimes. It's quite good. :-)

samarion:
Could you also add this optimization to the Stainless deconstructors at some point? Thanks!
Also, I noticed that you didn't change the way types are deconstructed, was there a reason for this?

mantognini:
> Could you also add this optimization to the Stainless deconstructors at some point? Thanks!

Sure, I'll test that and see if it helps as well.

> I noticed that you didn't change the way types are deconstructed, was there a reason for this?

Purely a gut feeling that with significantly less `case`s the bytecode might be more efficient with one less indirection. But we never know so I'll make some experiments.

PR #29, opened 2017-06-16
Allow specialised extractor to extend dispatch table

mantognini:
Further changes for faster extraction. This changes are required to allow the same optimisation in stainless.

### DIFF
--- a/src/main/scala/inox/ast/Extractors.scala
+++ b/src/main/scala/inox/ast/Extractors.scala
@@ -28,6 +28,8 @@ trait TreeDeconstructor {
   /** Extracted subtrees from an expression as well as a "builder" */
   type DeconstructedExpr = (Seq[s.Variable], Seq[s.Expr], Seq[s.Type], ExprBuilder)
 
+  type ExpressionTableDispatch = Map[Class[_], s.Expr => DeconstructedExpr]
+
   /** Optimisation trick for large pattern matching sequence: jumps directly to the correct case based
     * on the type of the expression -- a kind of virtual table for pattern matching.
     *
@@ -37,7 +39,10 @@ trait TreeDeconstructor {
     *
     *     classOf[s.Not] -> { case s.Not(e) => /* ... */ }
     */
-  private val exprTable: Map[Class[_], s.Expr => DeconstructedExpr] = HashMap(
+  private lazy val exprTable: ExpressionTableDispatch = buildExprTableDispatch
+
+  // This is overridable
+  protected def buildExprTableDispatch: ExpressionTableDispatch = HashMap(
     /* Unary operators */
     classOf[s.Not] -> { expr =>
       val s.Not(e) = expr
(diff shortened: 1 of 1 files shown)


samarion:
Could you please check the slowdown we would see by preserving the current
API and using the map with getOrElse(super.deconstruct) in subclasses ?
Ideally I'd prefer to keep a nice and clean API like that (although this
would be a bit faster indeed).

mantognini:
I had some problems reproducing my speedups, probably something to do with non-determinism in the solvers... so this might take a bit more time.

I think it makes sense to have only one overriding point, i.e. either the `exprTable` Map, or the `deconstruct` function. So, *if* we see a performance improvement using exclusively the Map in all sub-classes, then I'd recommend this patch. I'll let you know when I've got stable performance results.

samarion:
Sounds good. If you focus on timing the extraction suite, you'll probably
get more consistent results than with those involving solvers. The
extraction suite should use the deconstructors enough to enable comparison.

mantognini:
Here are some results:

Using `perf stat` to aggregate runtimes on 15 runs of `sbt 'stainless-dotty/it:testOnly *Extract*'` (using dotty here to have more to extract) gives 

image


where *before* refers to the versions of stainless/inox before any optimisation changes, *dispatch_inox* corresponds to the current *master* branches and *dispatch_full* corresponds to the *perf* branches of both inox & stainless. The average (*AVG (test)*) runtime of the extraction suite is measured using the wallclock (seconds), while the average (*AVG (perf)*) runtime of the full sbt command is measured in cpu time (seconds).

Overall, the speedup is not as good as the one I reported earlier. I think I was "unlucky" with the performance of the SMT solvers when I ran my benchmarks last week and got increasingly better results. Yet, we still gain several %, which is nice I think.

Profiling the code is hard given that all the profiling tools I've tried crash at some point or another. The results however seems to show that other `deconstruct` functions (e.g. for type) take less time and we should not be too worried about them at this point.

samarion:
Well, it's still some improvement. Did you try timing stainless perf
through a local deconstructor map (defining the class to deconstructor map
as a private field in the stainless deconstructor and using
super.deconstruct)?

I think a 5% speedup overall is pretty good, though as you say, it's
probably not critical at this time.

mantognini:
No, I haven't benchmark any alternative versions. I could, but I'm not sure it's worth the time spend on that since the optimisation is not as majestic as I first thought. My guess is that it could be even slower that the current *dispatch_inox* version, because the most frequent trees (i.e. `Variable`, ..) would need a longer path until they are identified and deconstructed. Having only one hash table ensures we find the matching dtor in one pass, instead of having several fallbacks to other hash tables / patmat. Do you think this gut feeling is correct enough to avoid investigating the speedup difference?

samarion:
I'm pretty sure the current version with the shared map will be faster, but
I don't like exposing this low level optimization API as the default API
for deconstructors. If the difference is not significant, I'd prefer to
keep the interface as is (override deconstruct) and let people write their
own optimizations if they want to. We would pay a small performance penalty
but would get a much nicer API I feel.

mantognini:
I understand. I'll investigate that further in the coming days then.

mantognini:
Closing for reasons exposed in epfl-lara/stainless#49.

PR #31, opened 2017-07-25
removes the simplification that replaces an expression by a variable

jad-hamza:
This is the case that replaces expression by variables. Related to https://github.com/epfl-lara/stainless/issues/62.


samarion:
Merged, thanks!

PR #33, opened 2017-08-07
Fix printer for Char literals

mantognini:
Obviously this is a minor fix.

Previously `'\n'` (and other escaped char) was rendered as a newline. Now it's rendered as expected, with single quotes.


samarion:
Merged, thanks!

PR #34, opened 2017-08-10
Adding traversal and transformations for Identifiers

jad-hamza:
I use this for the Canonization that normalizes Identifiers (https://github.com/jad-hamza/stainless/blob/d2ab0de37cc36511172e3224f5b4a88f3656293a/core/src/main/scala/stainless/transformers/Canonization.scala).
This is used for implementing the cache of verification conditions.

I may have forgotten identifiers in some places.

(I made the corresponding changes in Stainless as well.)


samarion:
Merged, thanks!

PR #35, opened 2017-08-19
Adds a helper method for changing the invariant of an ADT

jad-hamza:
This is needed when transforming programs in canonical form.


samarion:
This is way to dangerous, we could have inconsistent programs where the invariant is not part of the symbol table. This would crash the CodeGenEvaluator and probably cause lots of other suffering. The invariant is simply stored as a function in the Symbols table and indexed by the id, so you should be able to create consistent programs that maintain this (for lack of a better word) invariant.

jad-hamza:
Makes sense, thanks

PR #36, opened 2017-08-19
missing transformations and traversals for ids

jad-hamza:


src/main/scala/inox/ast/TreeOps.scala
```diff
@@ -77,7 +77,7 @@ trait TreeOps { self: Trees =>
       case sort: ADTSort =>
         traverse(sort.id)
         sort.tparams.foreach(traverse)
-        sort.cons.foreach(traverse)
+        sort.cons.sortBy(_.name).foreach(traverse)
```
samarion: This is already a Seq, so the traversal order is deterministic. This order is used pretty much everywhere so it would be a bit weird to change it here.

jad-hamza:
Ok thanks I'll reverse that. Can there be non determinism when the Seq is built in the first place?

samarion:
Yes, that's quite possible depending on when it was built. The non-determinism might even occur within the Scala compiler itself...

jad-hamza:
Would it be ok to sort by name at the moment the Seq is built?

samarion:
Sorting by name within the Scala and Dotty frontends would be fine. Hopefully that will make the whole system deterministic.

jad-hamza:
Ok I will try that thanks. I removed the reorderings from the transformations and traversals.

PR #38, opened 2017-09-05
add some missing positions

jad-hamza:
(This PR also changes the toString of Position's so that the File is displayed.)

### DIFF
--- a/src/main/scala/inox/ast/DSL.scala
+++ b/src/main/scala/inox/ast/DSL.scala
@@ -65,7 +65,7 @@ trait DSL {
 
     def getField(selector: Identifier) = ADTSelector(e, selector)
 
-    def apply(es: Expr*) = Application(e, es.toSeq)
+    def apply(es: Expr*) = Application(e, es.toSeq).copiedFrom(e)
 
     def isInstOf(tp: ADTType) = IsInstanceOf(e, tp)
     def asInstOf(tp: ADTType) = AsInstanceOf(e, tp)
--- a/src/main/scala/inox/ast/DSL.scala
+++ b/src/main/scala/inox/ast/DSL.scala
@@ -65,7 +65,7 @@ trait DSL {
 
     def getField(selector: Identifier) = ADTSelector(e, selector)
 
-    def apply(es: Expr*) = Application(e, es.toSeq).copiedFrom(e)
+    def apply(es: Expr*) = Application(e, es.toSeq)
 
     def isInstOf(tp: ADTType) = IsInstanceOf(e, tp)
     def asInstOf(tp: ADTType) = AsInstanceOf(e, tp)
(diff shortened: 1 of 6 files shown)


mantognini:
> (This PR also changes the toString of Position's so that the File is displayed.)

This is actually very good for reports. :+1:

src/main/scala/inox/ast/DSL.scala
```diff
@@ -65,7 +65,7 @@ trait DSL {
 
     def getField(selector: Identifier) = ADTSelector(e, selector)
 
-    def apply(es: Expr*) = Application(e, es.toSeq)
+    def apply(es: Expr*) = Application(e, es.toSeq).copiedFrom(e)
```
samarion: I'm not sure it makes much sense to assign the position directly within the DSL, especially in such a one-of way. It would be better if we could assign the position where the DSL is used.

jad-hamza: I removed this, looks like we still have no missing positions in the benchmarks

src/main/scala/inox/utils/Positions.scala
```diff
@@ -73,6 +73,8 @@ abstract class DefinedPosition extends Position {
 case class OffsetPosition(line: Int, col: Int, point: Int, file: File) extends DefinedPosition {
   def focusBegin = this
   def focusEnd = this
+
+  override def toString = super.toString + " (" + file + ")"
```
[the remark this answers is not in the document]
jad-hamza: Ah right, I removed this and now use the already defined fullString to display the positions in Stainless

src/main/scala/inox/ast/SymbolOps.scala
```diff
@@ -1364,7 +1364,7 @@ trait SymbolOps { self: TypeOps =>
 
     symbols.instantiation_>:(formalType, actualType) match {
       case Some(tmap) =>
-        FunctionInvocation(fd.id, fd.tparams map { tpd => tmap.getOrElse(tpd.tp, tpd.tp) }, args)
+        FunctionInvocation(fd.id, fd.tparams map { tpd => tmap.getOrElse(tpd.tp, tpd.tp) }, args).setPos(fd.getPos)
```
samarion: I hadn't noticed this but I don't think setting the position of the function to a call site makes much sense.
It would be better to call `copiedFrom` at the place where `functionInvocation` was called.

src/main/scala/inox/ast/SymbolOps.scala
```diff
@@ -1420,7 +1420,7 @@ trait SymbolOps { self: TypeOps =>
     val actualType = tupleTypeWrap(args.map(_.getType))
 
     symbols.instantiation_>:(formalType, actualType) match {
-      case Some(tmap) => ADT(instantiateType(adt.typed.toType, tmap).asInstanceOf[ADTType], args)
+      case Some(tmap) => ADT(instantiateType(adt.typed.toType, tmap).asInstanceOf[ADTType], args).setPos(adt.getPos)
```
samarion: Same remark as for the `functionInvocation` case.

samarion:
I merged and removed the two `setPos` I commented about, I still need to check whether we're now missing positions from the Stainless reports.

PR #39, opened 2017-09-22
Making Timer API thread safe AND exception safe

mantognini:
Fix timers, especially when using multithreading, and make code more robust toward undesired exceptions.


samarion:
LGTM. Thanks, this was also on my list since the weird times started appearing in reports!

mantognini:
You're welcome! I'll create a similar PR for stainless.

PR #40, opened 2017-09-22
Compute relative path for DefinedPosition

mantognini:
This makes path more consistent when printed, especially when loading caches.

Instead of printing sometimes a relative path (because the `File` was created with a relative path) or sometimes a full path (e.g. when a position is recovered from a cache, the `File` is reconstructed with an absolute path), with this patch all paths are printed with a relative path to the current working directory.


samarion:
LGTM.

PR #41, opened 2017-09-26
Improve consistency of command line names

mantognini:
Using `-` to separate words in command line options.

### DIFF
--- a/src/main/scala/inox/Model.scala
+++ b/src/main/scala/inox/Model.scala
@@ -2,7 +2,7 @@
 
 package inox
 
-object optPrintChooses extends FlagOptionDef("printchooses", false)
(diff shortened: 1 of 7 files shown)


samarion:
LGTM, but it seems we're still missing `optAssumeChecked` no?

mantognini:
Yes, indeed. I didn't noticed that one because `stainless --help` doesn't list it. [It's being removed from the options actually](https://github.com/epfl-lara/stainless/blob/5439772c19f4fe289f819b9206b8e7ac5a413394/core/src/main/scala/stainless/MainHelpers.scala#L41). I'll update it ASAP.

mantognini:
Done.

samarion:
Looks good to merge.

PR #42, opened 2017-09-27
Quick'n'dirty fix for SimplifierWithPC

mantognini:
This ensures that the trait is thread safe. This is required for the concurrency introduced by epfl-lara/stainless#76.

Ideally, it's internal state shouldn't be mutable and/or the critical section should be reduced to its minimal size for better throughput efficiency.

I couldn't provide a more efficient patch mainly because this feature is not well documented and I couldn't infer all the assumptions easily. This is therefore more a temporary workaround than a proper patch. I basically added  `synchronized` on all publicly available methods' body.

### DIFF
--- a/src/main/scala/inox/transformers/SimplifierWithPC.scala
+++ b/src/main/scala/inox/transformers/SimplifierWithPC.scala
@@ -177,24 +177,26 @@ trait SimplifierWithPC extends TransformerWithPC { self =>
 
   private val pureCache: MutableMap[Identifier, Boolean] = MutableMap.empty
 
-  private def isPureFunction(id: Identifier): Boolean = opts.assumeChecked || (pureCache.get(id) match {
-    case Some(b) => b
-    case None =>
-      val fd = getFunction(id)
-      if (transitivelyCalls(fd, fd)) {
-        pureCache += id -> false
-        false
-      } else {
-        pureCache += id -> true
-        if (isPure(fd.fullBody, CNFPath.empty)) {
-          true
-        } else {
+  private def isPureFunction(id: Identifier): Boolean = opts.assumeChecked || synchronized {
+    pureCache.get(id) match {
+      case Some(b) => b
+      case None =>
+        val fd = getFunction(id)
+        if (transitivelyCalls(fd, fd)) {
           pureCache += id -> false
-          transitiveCallers(fd).foreach(fd => pureCache += fd.id -> false)
           false
+        } else {
+          if (isPure(fd.fullBody, CNFPath.empty)) {
+            pureCache += id -> true
+            true
+          } else {
+            pureCache += id -> false
+            transitiveCallers(fd) foreach { pureCache += _.id -> false }
+            false
+          }
         }
-      }
-  })
+    }
+  }
 
   private def isInstanceOf(e: Expr, tpe: ADTType, path: CNFPath): Option[Boolean] = {
     val tadt = tpe.getADT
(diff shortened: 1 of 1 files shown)


samarion:
Unfortunately, simulating stack on heap is dirty and can't really be made concurrency friendly I believe. WDYT of relying on thread-local storage, would that solve the issue without such a strong requirement on synchronization?

mantognini:
I didn't think of this.. it works apparently. :+1:

samarion:
You probably still need to synchronize the `isPureFunction` method since it puts temporary values into the `pureCache` map.

src/main/scala/inox/transformers/SimplifierWithPC.scala
```diff
@@ -177,24 +177,26 @@ trait SimplifierWithPC extends TransformerWithPC { self =>
...
+        if (transitivelyCalls(fd, fd)) {
           pureCache += id -> false
-          transitiveCallers(fd).foreach(fd => pureCache += fd.id -> false)
           false
+        } else {
+          if (isPure(fd.fullBody, CNFPath.empty)) {
+            pureCache += id -> true
+            true
+          } else {
+            pureCache += id -> false
```
mantognini: Mind the fact I've moved this line inside the `isPure` branch in order to avoid inserting the wrong value when `fd` is not pure.

mantognini:
Performance wise, I haven't noticed a significant difference between the `synchronize`d version and the one based on thread local variables.

src/main/scala/inox/transformers/SimplifierWithPC.scala
samarion: Ah yes, it seems this was from the time when I wasn't considering recursion as impure.

A better version of this function would be something like:
```scala
sealed abstract class PurityCheck
case object Pure extends PurityCheck
case object Impure extends PurityCheck
case object Checking extends PurityCheck
private[this] val pureCache: MutableMap[Identifier, PurityCheck] = MutableMap.empty

private def isPureFunction(id: Identifier): Boolean =
  opts.assumeChecked ||
  synchronized(pureCache.get(id) match {
    case Some(Pure) => true
    case Some(Impure) => false
    case Some(Checking) =>
      // Function is recursive and cycles were not pruned by the simplifier.
      // No need to update pureCache here as the update will take place in the next branch.
      false
    case None =>
      pureCache.value += id -> Checking
      val p = isPure(getFunction(id).fullBody, CNFPath.empty)
      pureCache.value += id -> (if (p) Pure else Impure)
      p
  })
```
This way, we consider cycles that the simplifier couldn't prune to imply impurity but dead-code won't lead to impurity.

I believe it's best to use a synchronized block here as the results can be shared across threads. If you want to really only guard the necessary blocks, the synchronization can take place only over the `Some(Checking)` and `None` cases (as a single block).

mantognini: I'll add this code as-is. The `synchronized` block is fine I think -- my guess is that performance shouldn't change significantly and therefore shouldn't dictate a more complicated design.

mantognini: Done. Juste changed `pureCache.value` to `pureCache`. I guess it was just a minor typo, right? Or did you intend to use `DynamicVariable` here as well?

samarion: Yes, that was a typo, thanks. I started by writing a thread-local version but then realized this would disable result sharing. Looks good to merge.

PR #43, opened 2017-10-05
Fix a few typos in documentation

mantognini:
This is minor but I stumbled upon a duplicate of the documentation so I've removed it.

Commit history: [/doc](https://github.com/epfl-lara/inox/commits/master/doc) vs [/src/main/doc](https://github.com/epfl-lara/inox/commits/master/src/main/doc)
Diff between the directories: [gist](https://gist.github.com/mantognini/4c1c6cfbf74a33ff780451cbb33e3942) -- doesn't include corrections of typos.


samarion:
Aha these aren't duplicates, the docs under /doc are generated from those under /src/main/doc by tut (see build.sbt). This checks that our code examples can indeed be executed. Maybe we could add a comment to some of the files to indicate this.

But thanks for catching the typos! Please fix them in /src/main/doc, tut will copy the files over to /doc.

mantognini:
Don't you think we should remove from git any file that is generated by scripts?

Also, shouldn't those tests checked by CI?

samarion:
We can't remove them from git as the files in /doc _are_ the documentation. The annotations in /src/main/doc prevent the code to be rendered correctly by github so we can't present those files as the doc directly. We could push the generated docs to github wiki pages but given the small amount of documentation we have, I don't think it makes sense to host it in a different repo.

We aren't really testing anything, just making sure the code in the documentation actually compiles. It's quite common for APIs to change and the doc to then be out of sync, this approach gives us a signal when that occurs.

I agree that the duplication is not great, but I don't really see a better way of doing this. We would need github to provide us with a way to render code nicely _without_ using the scala tag.

mantognini:
I think it's cleaner to have the doc hosted on web server, this way we could remove the dups easily -- but that's just my personal taste. I'll fix the typos in the correct directory later. ;-)

As for the CI, do you want me to add `sbt tut`? Because currently nothing is tested I believe.

samarion:
I agree it's cleaner in general but also more work and more maintenance. I really like the fact that documentation and code can be kept at the same place in github (not really a fan of having to publish doc somewhere else when using sphinx, for example). What would be really great would be if we could tell github to render the `tut:silent` tags as scala code.

Tut is run during compilation, so if your project builds, then the doc has been tested. There might be some issues with dependency computations due to sbt though, so this might not always be true... Not sure it's worth adding more larabot checks for these (hopefully) rare occurrences.

mantognini:
> Tut is run during compilation

Oh, right, I missed this bit...

mantognini:
Here. I've also added a comment at the top of each file which should help people making the same mistake as I did ([syntax for comment](https://stackoverflow.com/a/20885980/520217)).

samarion:
LGTM! Thanks

PR #44, opened 2017-10-09
Ensure variables are available in the scope when used

mantognini:
This catches error such as this one, but hopefully many more:

```scala
  case class B(var x: Int) {
    def foo(y: Int): Boolean = {
      x == y // just a dummy example
    } ensuring { res =>
      old(x) == x // instead of old(this).x == x
    }
  }
```

Of course, we need to give the user a proper error message, but that's the job of stainless. This patch ensures all transformations don't mess (too much) with variables.

**BUT** there's a catch...

The TIP parser doesn't extract commands such as `(declare-const |error: Match is non-exhaustive!58| (_ BitVec 32))` into `Let` statement and the "variable" (`Parser.Locals.vars`) are not present in the `Symbols`, hence the tests are currently failing.

The question is, how do we workaround that? Should the TIP part be updated to use `Let`? (I've no idea how it works nor how much effort it would mean.) Should this kind of checks be disabled for some part of Inox? Should they be moved to stainless?

### DIFF
--- a/src/main/scala/inox/ast/Paths.scala
+++ b/src/main/scala/inox/ast/Paths.scala
@@ -11,19 +11,23 @@ trait Paths { self: SymbolOps with TypeOps =>
     /** Add a binding to this `PathLike`. */
     def withBinding(p: (ValDef, Expr)): Self
 
-    def withBindings(ps: Iterable[(ValDef, Expr)]): Self = ps.foldLeft(self)(_ withBinding _)
+    final def withBindings(ps: Iterable[(ValDef, Expr)]): Self = ps.foldLeft(self)(_ withBinding _)
+
+    def withBound(b: ValDef): Self
+
+    final def withBounds(bs: Iterable[ValDef]): Self = bs.foldLeft(self)(_ withBound _)
 
     /** Add a condition to this `PathLike`. */
     def withCond(e: Expr): Self
 
     /** Add multiple conditions to this `PathLike`. */
-    def withConds(es: Iterable[Expr]): Self = es.foldLeft(self)(_ withCond _)
+    final def withConds(es: Iterable[Expr]): Self = es.foldLeft(self)(_ withCond _)
 
     /** Appends `that` path at the end of `this`. */
     def merge(that: Self): Self
 
     /** Appends `those` paths at the end of `this`. */
-    def merge(those: Traversable[Self]): Self = those.foldLeft(self)(_ merge _)
+    final def merge(those: Traversable[Self]): Self = those.foldLeft(self)(_ merge _)
 
     /** Returns the negation of this path. */
     def negate: Self
@@ -36,21 +40,21 @@ trait Paths { self: SymbolOps with TypeOps =>
   implicit object Path extends PathProvider[Path] {
     final type Element = Either[(ValDef, Expr), Expr]
 
-    def empty: Path = new Path(Seq.empty)
+    def empty: Path = new Path(Seq.empty, Seq.empty)
 
     def apply(p: Expr): Path = p match {
-      case Let(i, e, b) => new Path(Seq(Left(i -> e))) merge apply(b)
-      case BooleanLiteral(true) => new Path(Seq.empty)
-      case _ => new Path(Seq(Right(p)))
+      case Let(i, e, b) => new Path(Seq(Left(i -> e)), Seq.empty) merge apply(b)
+      case BooleanLiteral(true) => new Path(Seq.empty, Seq.empty)
+      case _ => new Path(Seq(Right(p)), Seq.empty)
     }
 
-    def apply(p: (ValDef, Expr)): Path = new Path(Seq(Left(p)))
+    def apply(p: (ValDef, Expr)): Path = new Path(Seq(Left(p)), Seq.empty)
 
-    def apply(path: Seq[Expr]): Path = new Path(path.filterNot(_ == BooleanLiteral(true)).map(Right(_)))
+    def apply(path: Seq[Expr]): Path = new Path(path.filterNot(_ == BooleanLiteral(true)).map(Right(_)), Seq.empty)
   }
 
   /** Encodes path conditions
-    * 
+    *
     * Paths are encoded as an (ordered) series of let-bindings and boolean
     * propositions. A path is satisfiable iff all propositions are true
     * in the context of the provided let-bindings.
@@ -59,18 +63,22 @@ trait Paths { self: SymbolOps with TypeOps =>
     * not defined, whereas an encoding of let-bindings with equalities
     * could introduce non-sensical equations.
     */
-  class Path protected(val elements: Seq[Path.Element]) extends Printable with PathLike[Path] {
+  class Path protected(val elements: Seq[Path.Element], val bounds: Seq[ValDef])
+    extends Printable with PathLike[Path] {
     import Path.Element
 
     /** Add a binding to this [[Path]] */
-    def withBinding(p: (ValDef, Expr)) = {
+    override def withBinding(p: (ValDef, Expr)) = {
       def exprOf(e: Element) = e match { case Right(e) => e; case Left((_, e)) => e }
       val (before, after) = elements span (el => !exprOps.variablesOf(exprOf(el)).contains(p._1.toVariable))
-      new Path(before ++ Seq(Left(p)) ++ after)
+      new Path(before ++ Seq(Left(p)) ++ after, bounds)
     }
 
+    /** Add a bound to this [[Path]], a variable being defined but to an unknown/arbitrary value. */
+    override def withBound(b: ValDef) = new Path(elements, b +: bounds)
+
     /** Add a condition to this [[Path]] */
-    def withCond(e: Expr): Path = e match {
+    override def withCond(e: Expr): Path = e match {
       case TopLevelAnds(es) if es.size > 1 => withConds(es)
       case Not(TopLevelOrs(es)) if es.size > 1 => withConds(es.map(not(_)))
       case _ =>
@@ -81,16 +89,16 @@ trait Paths { self: SymbolOps with TypeOps =>
           exprOps.replace(Map(not(c) -> BooleanLiteral(false)), e)
         }))
 
-        new Path(newElements.filterNot(_.right.exists(_ == BooleanLiteral(true))))
+        new Path(newElements.filterNot(_.right.exists(_ == BooleanLiteral(true))), bounds)
     }
 
     /** Remove bound variables from this [[Path]]
       * @param ids the bound variables to remove
       */
-    def --(vds: Set[ValDef]) = new Path(elements.filterNot(_.left.exists(p => vds(p._1))))
+    def --(vds: Set[ValDef]) = new Path(elements.filterNot(_.left.exists(p => vds(p._1))), bounds filterNot vds)
 
     /** Appends `that` path at the end of `this` */
-    def merge(that: Path): Path = new Path((elements ++ that.elements).distinct)
+    override def merge(that: Path): Path = new Path((elements ++ that.elements).distinct, that.bounds ++ bounds)
 
     /** Transforms all expressions inside the path
       *
@@ -100,10 +108,10 @@ trait Paths { self: SymbolOps with TypeOps =>
       * @see [[map(fVal:Paths\.this\.trees\.ValDef=>Paths\.this\.trees\.ValDef,fExpr:Paths\.this\.trees\.Expr=>Paths\.this\.trees\.Expr):Paths\.this\.Path* map]]
       *      for a map that can transform bindings as well
       */
-    def map(f: Expr => Expr) = new Path(elements.map(_.left.map { case (vd, e) => vd -> f(e) }.right.map(f)))
+    def map(f: Expr => Expr) = new Path(elements.map(_.left.map { case (vd, e) => vd -> f(e) }.right.map(f)), bounds)
 
     /** Transforms both let bindings and expressions inside the path
-      * 
+      *
       * The function `fVal` is applied to all values in [[bound]] and `fExpr` is applied
       * to both the bodies of the [[bindings]] as well as the [[conditions]].
       *
@@ -111,7 +119,8 @@ trait Paths { self: SymbolOps with TypeOps =>
       *      for a map defined only on expressions
       */
     def map(fVal: ValDef => ValDef, fExpr: Expr => Expr) = new Path(
-      elements.map(_.left.map { case (vd, e) => fVal(vd) -> fExpr(e) }.right.map(fExpr))
+      elements.map(_.left.map { case (vd, e) => fVal(vd) -> fExpr(e) }.right.map(fExpr)),
+      bounds
     )
 
     /** Instantiates type parameters within the path
@@ -123,7 +132,7 @@ trait Paths { self: SymbolOps with TypeOps =>
       val t = new TypeInstantiator(tps)
       new Path(elements.map(_.left.map {
         case (vd, e) => t.transform(vd) -> t.transform(e)
-      }.right.map(t.transform)))
+      }.right.map(t.transform)), bounds)
     }
 
     /** Splits the path on predicate `p`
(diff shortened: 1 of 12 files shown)


samarion:
I think this is a good feature to have.

For the match exhaustivity errors, these come from the old translation of match expressions from Stainless to Inox and shouldn't appear anymore. The issue can be fixed by replacing the use place of these free variables by chooses (not sure about the exact TIP syntax, some tip benchmarks should already use chooses). Note that you can't fix this with lets since these constants are used to encode non-determinism (that's why a choose will do the trick).

About the concrete implementation: please don't use `preMapWithContext`. It doesn't understand the precise variable binding rules and can lead to inconsistencies when used for variable scoping. For example, given `Let(i, e, b)`, it will think `i` is bound in `e`. You can use the transformer with PC to get correct scoping rules for bindings or reimplement a precise traversal.

While you're at it, could you also add a check that makes sure the identifier uniquely determines both the type and flags of a variable? This should also be an invariant of the transformations.

Thanks!

mantognini:
> The issue can be fixed by replacing the use place of these free variables by chooses

I'm really not familiar with this TIP, would you mind giving me a hint on how to do this? Thanks.

> You can use the transformer with PC to get correct scoping rules for bindings or reimplement a precise traversal.

I'll need to implement a new visitor for that actually: in `TransformerWithPC`, the `if-else`, `and`, `or` and `implies` case are not appropriately handled for scope purposes because sibling expressions (e.g. `e1 && e2`) are handled as parent-children (e.g. `e2` is a child of `e1`).

> While you're at it, could you also add a check that makes sure the identifier uniquely determines both the type and flags of a variable?

Sure!

samarion:
There are two orthogonal path elements that `TransformerWithPC` is taking care of: path conditions and bindings. The children relation you're talking of is the path condition one (and `e2` should indeed depend on `e1` in `e1 && e2`). However, the bindings are handled separately and only appear when `path.withBinding` is called (so not in `if-else`, `and`, `or`, etc. If there is an issue with the scope handling of `TransformerWithPC`, then this is a bug that should be fixed!

The best interface into `TransformerWithPC` is through `SymbolOps.transformerWithPC` as this will work with tree extensions in Stainless. There are a few examples in `SymbolOps` of how to use it.
Reimplementing the traversal is of course also an option, but it won't be compatible with tree extensions.

For the chooses, see https://github.com/epfl-lara/inox/blob/f105308eb60913349911457b4aeaa0c5ff65d485/src/it/resources/regression/tip/UNSAT/Tutorial.scala-1.tip#L21 (not easy to read, sorry).

It basically boils down to replacing
```
(declare-const |error: Match is non-exhaustive!27| SomeType)
...
[...] |error: Match is non-exhaustive!27| [...]
```
pairs with
```
[...] (choose |error: Match is non-exhaustive!27| SomeType true) [...]
```

mantognini:
> There are two orthogonal path elements that TransformerWithPC is taking care of: path conditions and bindings.

I guess I missed some bit of information when reading the implementation. Will try with that. ;-)

Thanks for the example about choose. Should I update the test ressources to use `choose`, or should I add some transformation to go from abstract constant to choose statement? Are there some limitation to take into account if I implement some transformation? (e.g. can the same constant be use in two places and therefore be translated to two chooses?)

samarion:
Hmm, I believe constants are allowed to be type-parametric in TIP, so the ideal solution is probably to parse `declare-const` as a FunDef with a choose body in the TIP parser. Allowing global variables is not great since these can't be type-parametric, so I think it would be best to completely ban them.

samarion:
Ah, it seems I forgot about variables that are introduced by lambdas, foralls, chooses, etc.
I'm not sure what the best way to handle these is...

I guess it could be useful to extend the `TransformerWithPC` to keep track of non-free variables. This can be done by extending the `PathLike` interface with `withBound` or something and adding the relevant cases to `TransformerWithPC`.

mantognini:
Yes, I started writing about that before lunch but got interrupted. Also, I haven't worked on the TIP part yet, because there were several issues with `collectWithPC` that were more pressing.

The first of which is that `Forall(args, body)` doesn't expand the set of bindings, as you've noticed. I would like to add something like this to the transformer with PC:

```scala
    case Forall(args, b) =>
      val sb = rec(b, env withBindings args)
      Forall(args, sb).copiedFrom(e)
```

But this doesn't work because we need an expression to which the binding should be attached and `NoTree` isn't defined in inox. I could add `withBound` to workaround this, as you suggested.

The second issue is with stainless extensions, such as `LetVar`, which require the transformer to be specialised to handled them. And this is quite a nightmare I've to say. I've finally managed to get stainless to compile & use a custom `TransformerWithPC` for `imperative`, but it took me a really long time to get the inheritance/override right... Now I can look into actually implementing `rec`. :-)

samarion:
Yes, I agree the transformer inheritance mess is quite an issue, but I didn't have any better idea of how to allow `rec` being overriden by a closure. If you have a better idea / better design, please feel free to fix the ugly hacks I had to resort to! Having generic transformations that can be cleanly extended to work on different trees is quite hard it turns out...

src/main/scala/inox/ast/Paths.scala
```diff
@@ -59,18 +63,22 @@ trait Paths { self: SymbolOps with TypeOps =>
     * not defined, whereas an encoding of let-bindings with equalities
     * could introduce non-sensical equations.
     */
-  class Path protected(val elements: Seq[Path.Element]) extends Printable with PathLike[Path] {
+  class Path protected(val elements: Seq[Path.Element], val bounds: Seq[ValDef])
```
mantognini: Not sure if this is what `Path` was designed to become...

src/main/scala/inox/transformers/SimplifierWithPC.scala
```diff
@@ -132,7 +132,9 @@ trait SimplifierWithPC extends TransformerWithPC { self =>
       }
     }
 
-    def withCond(e: Expr) = if (formulaSize(e) > 20) this else {
+    override def withBound(b: ValDef) = ???
```
mantognini: I don't know how to implement this. But it doesn't seems to bother runtime, at least not in inox test suite.

src/main/scala/inox/tip/Parser.scala
```diff
@@ -142,13 +142,22 @@ class Parser(file: File) {
...
       val tpsLocals = locals.withGenerics(tps.map(s => s -> TypeParameter.fresh(s.name).setPos(s.optPos)))
-      (None, locals.withVariable(sym,
-        Variable.fresh(sym.name, tpsLocals.extractSort(sort)).setPos(sym.optPos)))
+      val choose = Choose(
+        ValDef(FreshIdentifier(sym.name), tpsLocals.extractSort(sort)),
+        BooleanLiteral(true)
+      ).setPos(sym.optPos)
+      (None, locals.withVariable(sym, choose)) // FIXME This failes:
+// [info] -  77: UNSAT - MergeSort2.scala-0.tip solver=smt-z3 assumechecked *** FAILED *** (128 milliseconds)
+// [info]   java.lang.ClassCastException: inox.ast.Expressions$Choose cannot be cast to inox.ast.Expressions$Variable
```
mantognini: It seems doing this for parametrised (declare const ...) is not okay.

src/main/scala/inox/tip/Parser.scala
```diff
@@ -142,13 +142,22 @@ class Parser(file: File) {
       (Some(tpsLocals.extractTerm(term)), locals)
 
     case DeclareConst(sym, sort) =>
-      (None, locals.withVariable(sym,
-        Variable.fresh(sym.name, locals.extractSort(sort)).setPos(sym.optPos)))
+      val choose = Choose(
+        ValDef(FreshIdentifier(sym.name), locals.extractSort(sort)),
+        BooleanLiteral(true)
+      ).setPos(sym.optPos)
+      (None, locals.withVariable(sym, choose))
```
mantognini: Does this correspond to what you had in mind?

src/main/scala/inox/ast/Paths.scala
samarion: I think it's not a bad thing to know about bound variables in the path. It could be useful for other things than well-formedness checks at some point. However, I think it would be better if you put the bounds inside the `elements` list as well so we can preserve the order of bindings/conditions/bounds.

It's probably not a good idea to use `Either` for such things anyway, they make the code hard to read...

src/main/scala/inox/tip/Parser.scala
samarion: I think I would go a `FunDef` instead of variable, to avoid repeating the choose for multiple uses. Something like the `DeclareFun` case.

src/main/scala/inox/tip/Parser.scala
samarion: Basing yourself off `DeclareFunPar` will fix this.

src/main/scala/inox/transformers/SimplifierWithPC.scala
samarion: We don't need to keep track of bound variables in `CNFPath`, but please return `this` instead of `???` to avoid crashing if we encounter a `Forall` or something.

mantognini: done.

src/main/scala/inox/ast/Paths.scala
mantognini: I'll look into this next week

src/main/scala/inox/tip/Parser.scala
mantognini: I've updated the code as you suggested but now I've got some other issues. Some test are no longer SAT and some seems to enter an infinite loop or something. Maybe you have some insight on the issue here?

samarion: Ah, I forgot that free variables in the final expression we want to check are also output as `declare-const` when printing TIP programs... So we should only use the `DeclareFun` trick in the type parametric case and rewrite the benchmarks with `declare-fun` instead of `declare-const` for the match exhaustiveness variables.

However, I don't see any reason for the benchmarks to fail with chooses instead of variables, so it would be useful to debug this anyway!

mantognini: I tried to make the changes but either I did something slightly wrong or there's another bug because it:test gets stuck...

mantognini:
I've reverted changes to the parser, instead we should modify the output so that function declarations are used. Unless you see what was the issue making the tests loop forever and you want to fix it, of course.

samarion:
I pushed 09b37f648089df39dd4187c681d040bdfe801c68 which partially fixes the issue with choose evaluation. There are still testcases that don't pass with `declare-const` parsed as a `declare-fun` but these are due to the quantifier instantiation heuristics that don't handle variables exactly the same way as function invocations. I could probably make them work in the specific case of free functions but I don't think there's much point.

I agree with reverting the parser changes though, I think it's best to change the output, as you said. We print variables as `declare-const`, so it makes sense to parse them from `declare-const` as well.

mantognini:
I'm not sure I've understood it right, so just to be sure, do you agree with those changes: 
4fec121? i.e. applying these commands

    sed -i.bak -e 's/(declare-const |\(error: [^|]*\)| /(declare-fun |\1| () /' src/it/resources/regression/tip/{PARSING/,SAT/,UNKNOWN/,UNSAT/}*.tip
    sed -i.bak -e 's/(declare-const (par \((.*)\) (|\(error: [^|]*\)| /(declare-fun (par \1 (|\2| () /' src/it/resources/regression/tip/{PARSING/,SAT/,UNKNOWN/,UNSAT/}*.tip

(The missing bit is of course to output declare-fun for pat-mat errors.)

samarion:
Yes, I agree with the sed changes.
And pat-mat errors are already output as chooses by the current version of Stainless. These tip benchmarks were generated by an older version of Stainless.

mantognini:
The last commit and the corresponding one in epfl-lara/stainless#89 update the `Path` implementation to use explicit types instead of `Either`. Let me know what you thing of it, especially where I've put `FIXME`s. :-)

src/main/scala/inox/ast/Paths.scala
```diff
@@ -63,52 +70,53 @@ trait Paths { self: SymbolOps with TypeOps =>
...
+        case CloseBound(_, e) => e
+        case Condition(e) => e
+      }
+      assert(exprs forall { e => !(exprOps.variablesOf(e) contains vd.toVariable) })
+      this :+ CloseBound(vd, value)
     }
 
     /** Add a bound to this [[Path]], a variable being defined but to an unknown/arbitrary value. */
-    override def withBound(b: ValDef) = new Path(elements, b +: bounds)
+    override def withBound(b: ValDef) = this :+ OpenBound(b)
```
samarion: Maybe have the same assert as for `withBinding`?

src/main/scala/inox/ast/Paths.scala
```diff
@@ -185,27 +181,37 @@ trait Paths { self: SymbolOps with TypeOps =>
...
-      new Path(newElements, bounds)
+
+      new Path(newElements)
     }
 
     /** Free variables within the path */
-    lazy val variables: Set[Variable] = fold[Set[Variable]](Set.empty,
-      (vd, e, res) => res - vd.toVariable ++ exprOps.variablesOf(e), (e, res) => res ++ exprOps.variablesOf(e)
-    )(elements)
+    lazy val freeVariables: Set[Variable] = {
```
samarion: I like the name `variables`, it fits with `variablesOf`, but I understand if you want to keep `freeVariables`.

src/main/scala/inox/ast/Paths.scala
```diff
@@ -214,8 +220,9 @@ trait Paths { self: SymbolOps with TypeOps =>
       */
     private def fold[T](base: T, combineLet: (ValDef, Expr, T) => T, combineCond: (Expr, T) => T)
                        (elems: Seq[Element]): T = elems.foldRight(base) {
-      case (Left((id, e)), res) => combineLet(id, e, res)
-      case (Right(e), res) => combineCond(e, res)
+      case (CloseBound(vd, e), res) => combineLet(vd, e, res)
+      case (Condition(e), res) => combineCond(e, res)
+      case (OpenBound(_), res) => res // FIXME should it also take a combiner for OpenBound?
```
samarion: I don't think `OpenBound` can be combined in any useful way. If we come up with a use-case later on, it'll be easy to extend the `fold` function at that time.

src/main/scala/inox/ast/SymbolOps.scala
```diff
@@ -1053,6 +1054,7 @@ trait SymbolOps { self: TypeOps =>
             nl
 
           case fi: FunctionInvocation =>
+            // FIXME should this also use "open" bounds?
```
samarion: No, I don't think so. We are only interested in let bindings where the body (transitively) contains a function invocation, which obviously isn't possible for an open bound.

src/main/scala/inox/ast/SymbolOps.scala
```diff
@@ -1073,7 +1075,7 @@ trait SymbolOps { self: TypeOps =>
 
       def replace(path: Path, oldE: Expr, newE: Expr, body: Expr): Expr = transformWithPC(body) {
         (e, env, op) =>
-          if ((path.bindings.toSet subsetOf env.bindings.toSet) &&
+          if ((path.bindings.toSet subsetOf env.bindings.toSet) && // FIXME bindings or bounds?
```
samarion: Probably both. You can probably add the condition `path.bounds subsetOf env.bounds` as well.

samarion:
I found the issue and fixed it. I'll push as soon as I've had time to run the Stainless regressions off the new version.

mantognini:
I've addressed the above comments. Let me know if there's something more.

src/main/scala/inox/ast/Paths.scala
mantognini: I've kept `freeVariables` because it's confusing otherwise. ;-)

samarion:
Most of this has been merged except for last two commits (I had too many changes to cherry-pick them once they arrived, sorry). You can port the difference if you want but the code should already be quite similar.

PR #45, opened 2017-10-11
Improve parsing of TIP files, add regression suite

mantognini:
See discussion in epfl-lara/stainless#81. I just fixed the one I've found, more cases might still be missing, but at least the tip files generated for stainless valid verification suite are all successfully parsed.


src/it/scala/inox/tip/TipParsingSuite.scala
```diff
@@ -0,0 +1,13 @@
+/* Copyright 2009-2016 EPFL, Lausanne */
+
+package inox
+package tip
+
+class TipParsingSuite extends TestSuite with ResourceUtils {
+  for (file <- resourceFiles("regression/tip/PARSING", _.endsWith(".tip"))) {
+    test(s"PARSING - ${file.getName}") { ctx =>
+      new Parser(file).parseScript
```
samarion: I think it would be best to also check whether the parsed script typechecks.
You can plug into the TipPrintingSuite with the PARSING sources, that way you'll get a more comprehensive test. Also, you should maybe change the TipPrintingSuite to call `ensureWellFormed` instead of the ad-hoc type-checking code currently used.

mantognini:
Here, the latest commit should improve the test suite further.

samarion:
LGTM

PR #48, opened 2017-11-13
[WIP] Add a ground simplifier

romac:
This simplifier traverse the given expression, and evaluate all its pure sub-expressions, given that they don't refer to a quantifier, or a `choose` construct.

### DIFF
--- a/src/main/scala/inox/ast/SymbolOps.scala
+++ b/src/main/scala/inox/ast/SymbolOps.scala
@@ -115,19 +115,18 @@ trait SymbolOps { self: TypeOps =>
     case _ => false
   }
 
-  /** Returns true when the given expression is ground, pure, and does not contain
-   *  choose or quantifiers, and can thus be simplified.
-   */
-  def canBeForced(expr: Expr): Boolean = {
-    isGround(expr) && isPure(expr)(PurityOptions.AssumeChecked) && !containsChooseOrQuantifiers(expr)
-  }
-
   /** Simplify all the pure and ground sub-expressions of the given expression,
-   * by forcing them using [[inox.evaluators.Evaluator]].
-   *
-   * @see [[canBeForced]]
-   */
-  def simplifyGround(expr: Expr)(implicit sem: symbols.Semantics, ctx: Context): Expr = {
+    * by evaluating them using [[inox.evaluators.Evaluator]].
+    * If `force` is omitted, the given expression will only be evaluated if it is
+    * ground, pure, and does not contain choose or quantifiers.
+    */
+  def simplifyGround(expr: Expr, force: Boolean = false)
+                    (implicit sem: symbols.Semantics, ctx: Context, opts: PurityOptions): Expr = {
+
+    val canEvaluate: Boolean = force || {
+      isGround(expr) && isPure(expr) && !containsChooseOrQuantifiers(expr)
+    }
+
     val evaluator = sem.getEvaluator
 
     def err(desc: String, msg: String): Unit = {
@@ -136,21 +135,23 @@ trait SymbolOps { self: TypeOps =>
 
     import inox.evaluators.EvaluationResults._
 
-    postMap {
-      case e if canBeForced(e) =>
-        evaluator.eval(e) match {
-          case Successful(value) =>
-            Some(value)
-          case RuntimeError(msg) =>
-            err("runtime error", msg)
-            None
-          case EvaluatorError(msg) =>
-            err("evaluator error", msg)
-            None
-        }
+    val result = if (canEvaluate) {
+      evaluator.eval(expr) match {
+        case Successful(value) =>
+          Some(value)
+        case RuntimeError(msg) =>
+          err("runtime error", msg)
+          None
+        case EvaluatorError(msg) =>
+          err("evaluator error", msg)
+          None
+      }
+    }
+    else {
+      None
+    }
 
-      case _ => None
-    } (expr)
+    result.getOrElse(expr)
   }
 
   private val typedIds: MutableMap[Type, List[Identifier]] =
@@ -1160,12 +1161,13 @@ trait SymbolOps { self: TypeOps =>
     mergeCalls(liftCalls(expr))
   }
 
-  private[inox] def simplifyFormula(e: Expr)(implicit ctx: Context): Expr = {
+  private[inox] def simplifyFormula(e: Expr)(implicit ctx: Context, sem: symbols.Semantics): Expr = {
     implicit val simpOpts = SimplificationOptions(ctx)
     implicit val purityOpts = PurityOptions(ctx)
 
     if (simpOpts.simplify) {
       val simp: Expr => Expr =
+        ((e: Expr) => simplifyGround(e))      compose
         ((e: Expr) => simplifyHOFunctions(e)) compose
         ((e: Expr) => simplifyExpr(e))        compose
         ((e: Expr) => simplifyForalls(e))     compose
--- a/src/main/scala/inox/ast/SymbolOps.scala
+++ b/src/main/scala/inox/ast/SymbolOps.scala
@@ -5,6 +5,7 @@ package ast
 
 import utils._
 import solvers.{PurityOptions, SimplificationOptions}
+import evaluators.EvaluationResults._
 
 import scala.collection.mutable.{Map => MutableMap, Set => MutableSet}
 
@@ -121,37 +122,24 @@ trait SymbolOps { self: TypeOps =>
     * ground, pure, and does not contain choose or quantifiers.
     */
   def simplifyGround(expr: Expr, force: Boolean = false)
-                    (implicit sem: symbols.Semantics, ctx: Context, opts: PurityOptions): Expr = {
+                    (implicit sem: symbols.Semantics, ctx: Context, opts: PurityOptions): Result[Expr] = {
+    import evaluators.optCrashOnChoose
 
-    val canEvaluate: Boolean = force || {
-      isGround(expr) && isPure(expr) && !containsChooseOrQuantifiers(expr)
+    val evalCtx = ctx.withOpts(optCrashOnChoose(true))
+    val evaluator = sem.getEvaluator(ctx)
+    val canEvaluate = force || isGround(expr) && isPure(expr)
+
+    if (canEvaluate) {
+      evaluator.eval(expr)
+    } else {
+      Successful(expr)
     }
+  }
 
-    val evaluator = sem.getEvaluator
-
-    def err(desc: String, msg: String): Unit = {
-      ctx.reporter.error(s"Forcing failed due to a $desc: $msg")
-    }
-
-    import inox.evaluators.EvaluationResults._
-
-    val result = if (canEvaluate) {
-      evaluator.eval(expr) match {
-        case Successful(value) =>
-          Some(value)
-        case RuntimeError(msg) =>
-          err("runtime error", msg)
-          None
-        case EvaluatorError(msg) =>
-          err("evaluator error", msg)
-          None
-      }
-    }
-    else {
-      None
-    }
-
-    result.getOrElse(expr)
+  private def simplifyGroundOrSame
+    (expr: Expr, force: Boolean = false)
+    (implicit sem: symbols.Semantics, ctx: Context, opts: PurityOptions): Expr = {
+    simplifyGround(expr, force).result.getOrElse(expr)
   }
 
   private val typedIds: MutableMap[Type, List[Identifier]] =
@@ -1167,11 +1155,11 @@ trait SymbolOps { self: TypeOps =>
 
     if (simpOpts.simplify) {
       val simp: Expr => Expr =
-        ((e: Expr) => simplifyGround(e))      compose
-        ((e: Expr) => simplifyHOFunctions(e)) compose
-        ((e: Expr) => simplifyExpr(e))        compose
-        ((e: Expr) => simplifyForalls(e))     compose
-        ((e: Expr) => simplifyAssumptions(e)) compose
+        ((e: Expr) => simplifyGroundOrSame(e)) compose
+        ((e: Expr) => simplifyHOFunctions(e))  compose
+        ((e: Expr) => simplifyExpr(e))         compose
+        ((e: Expr) => simplifyForalls(e))      compose
+        ((e: Expr) => simplifyAssumptions(e))  compose
         ((e: Expr) => mergeFunctions(e))
       simp(e)
     } else {
--- a/src/main/scala/inox/ast/SymbolOps.scala
+++ b/src/main/scala/inox/ast/SymbolOps.scala
@@ -115,6 +115,44 @@ trait SymbolOps { self: TypeOps =>
     case _ => false
   }
 
+  /** Returns true when the given expression is ground, pure, and does not contain
+   *  choose or quantifiers, and can thus be simplified.
+   */
+  def canBeForced(expr: Expr): Boolean = {
+    isGround(expr) && isPure(expr)(PurityOptions.AssumeChecked) && !containsChooseOrQuantifiers(expr)
+  }
+
+  /** Simplify all the pure and ground sub-expressions of the given expression,
+   * by forcing them using [[inox.evaluators.Evaluator]].
+   *
+   * @see [[canBeForced]]
+   */
+  def simplifyGround(expr: Expr)(implicit sem: symbols.Semantics, ctx: Context): Expr = {
+    val evaluator = sem.getEvaluator
+
+    def err(desc: String, msg: String): Unit = {
+      ctx.reporter.error(s"Forcing failed due to a $desc: $msg")
+    }
+
+    import inox.evaluators.EvaluationResults._
+
+    postMap {
+      case e if canBeForced(e) =>
+        evaluator.eval(e) match {
+          case Successful(value) =>
+            Some(value)
+          case RuntimeError(msg) =>
+            err("runtime error", msg)
+            None
+          case EvaluatorError(msg) =>
+            err("evaluator error", msg)
+            None
+        }
+
+      case _ => None
+    } (expr)
+  }
+
   private val typedIds: MutableMap[Type, List[Identifier]] =
     MutableMap.empty.withDefaultValue(List.empty)
 
--- a/src/main/scala/inox/ast/SymbolOps.scala
+++ b/src/main/scala/inox/ast/SymbolOps.scala
@@ -144,10 +144,11 @@ trait SymbolOps { self: TypeOps =>
       case Operator(es, recons) => recons(es.map(rec))
     }
 
+    // TODO: Should we run this within a fixpoint with simplifyByConstructor?
     def rec(e: Expr): Expr = e match {
       case e if isValue(e) =>
         e
-      case e if force || isPure(e) && isGround(e) =>
+      case e if isGround(e) && (force || isPure(e)) =>
         val evaluated = evaluator.eval(e)
         evaluated.result.getOrElse {
           if (force) ctx.reporter.error(SimplifyGroundError(e, evaluated))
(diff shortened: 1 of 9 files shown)


jad-hamza:
So @samarion, do you think we should rework the simplifier, and try to replace it by something similar to `simpl` based on computation rules? I guess that can be for a future PR. By the way, is there an evaluator in Stainless or Inox that implements the reduction/computation rules for terms with free variables?

src/main/scala/inox/ast/SymbolOps.scala
```diff
@@ -115,6 +115,44 @@ trait SymbolOps { self: TypeOps =>
     case _ => false
   }
 
+  /** Returns true when the given expression is ground, pure, and does not contain
+   *  choose or quantifiers, and can thus be simplified.
+   */
+  def canBeForced(expr: Expr): Boolean = {
+    isGround(expr) && isPure(expr)(PurityOptions.AssumeChecked) && !containsChooseOrQuantifiers(expr)
```
samarion: Note that you have to be careful about transitive function calls that contain chooses or quantifiers. A more precise approach would be to add an option to the evaluators to crash when encountering such an expression. Otherwise, you can compute the fixpoint of all transitive callees that contain chooses or quantifiers, but you will probably miss many opportunities to simplify in doing so since you'll have to consider code that wouldn't be evaluated given the concrete ground arguments.

Also, you can't use `AssumeChecked` here if it hasn't been enabled in the context. It would be better to take an implicit `PurityOptions` argument like `simplifyExpr`.

On a general design note, I would not declare this function outside of `simplifyGround`. The symbols get imported in many places so we're trying to avoid polluting the namespace unless it's either a useful override point or a useful standalone function.

src/main/scala/inox/ast/SymbolOps.scala
```diff
@@ -115,6 +115,44 @@ trait SymbolOps { self: TypeOps =>
...
+  def simplifyGround(expr: Expr)(implicit sem: symbols.Semantics, ctx: Context): Expr = {
+    val evaluator = sem.getEvaluator
+
+    def err(desc: String, msg: String): Unit = {
+      ctx.reporter.error(s"Forcing failed due to a $desc: $msg")
+    }
+
+    import inox.evaluators.EvaluationResults._
+
+    postMap {
```
samarion: Using a postMap here is valid but you will be invoking the evaluator many more times than necessary. For the `RecursiveEvaluator`, this probably doesn't matter much but with the `CodeGenEvaluator`, you get a non-trivial setup time (although evaluation is then much faster). I would use a dedicated recursive function that doesn't call itself on the evaluation result.

src/main/scala/inox/ast/SymbolOps.scala
```diff
@@ -115,6 +115,44 @@ trait SymbolOps { self: TypeOps =>
...
+      case e if canBeForced(e) =>
+        evaluator.eval(e) match {
+          case Successful(value) =>
+            Some(value)
+          case RuntimeError(msg) =>
+            err("runtime error", msg)
```
samarion: If we run the ground simplifier on every expression, this doesn't make much sense anymore :) The error reporting should be moved back into Stainless when evaluating calls to functions with the force annotation.

samarion:
@jad-hamza: I don't think we should replace the current simplifier (at least for now), but rather build the `simpl`-based simplifier and compare the two / run them both. I have a feeling `simpl` will be better suited to simplifying user code and the current simplifier will be better at eliminating code introduced by the transformation pipeline, but this is just a gut feeling and we should evaluate.

There is no evaluator in Inox based on expansion of non-ground terms. However, I expect it to look something like
```scala
def rec(e: Expr): Expr = e match {
  case fi @ FunctionInvocation(_, _, args) => rec(fi.tfd.paramSubst(fi.tfd.fullBody, args))
  case Let(i, e, b) => rec(replaceFromSymbols(Map(i -> e), b))
  case Operator(es, recons) => recons(es.map(rec))
}
```
with some extra simplification code for cases that are actually reducible. Also something to make it terminate :)

As the approaches seem relatively disjoint between the ground simplifier and `simpl` simplifier, it would be worth opening a new PR I think.

jad-hamza:
Sounds good thanks

src/main/scala/inox/ast/SymbolOps.scala
romac: Yes I figured that this is not the optimal way to do such things. Should the transformation be top-down or bottom-up then? Can you sketch (in a few words, no need for code except if it's easier for you to explain it that way), how such a recursive function would look like?  Something like ?

```scala
def rec(e: Expr): Expr = e match {
  case fi @ FunctionInvocation(_, _, args) => ???
  case Let(i, e, b) => ???
  case Operator(es, recons) => ???
}
```

src/main/scala/inox/ast/SymbolOps.scala
romac: > Note that you have to be careful about transitive function calls that contain chooses or quantifiers. A more precise approach would be to add an option to the evaluators to crash when encountering such an expression. Otherwise, you can compute the fixpoint of all transitive callees that contain chooses or quantifiers, but you will probably miss many opportunities to simplify in doing so since you'll have to consider code that wouldn't be evaluated given the concrete ground arguments.

Ok, I will go with the evaluator option then. Thanks.

> Also, you can't use AssumeChecked here if it hasn't been enabled in the context. It would be better to take an implicit PurityOptions argument like simplifyExpr.

Right, fixed! (commits incoming)

> On a general design note, I would not declare this function outside of simplifyGround. The symbols get imported in many places so we're trying to avoid polluting the namespace unless it's either a useful override point or a useful standalone function.

Of course, fixed as well.

src/main/scala/inox/ast/SymbolOps.scala
romac: Right, fixed! (commit incoming)

src/main/scala/inox/ast/SymbolOps.scala
samarion: Top down I'd say. As soon as you find something that is pure and ground, I'd try to simplify it. If the simplification fails, then recurse into the children. If it succeeds, then you have a value and are done. You can add some more checks if you want to avoid calling the evaluator when it's obvious it will fail but I'd just start with this for now and see whether you get a slowdown on existing benchmarks. I don't think we have so many ground expressions so it probably won't slow things down much. You can also avoid calling the simplifier if the expression is already a value.

So the cases in your function would be something like:
```scala
case e if isValue(e) => e
case e if isPure(e) && isGround(e) => // eval
case Operator(es, recons) => recons(es.map(rec))
```

src/main/scala/inox/evaluators/Evaluator.scala
```diff
@@ -4,6 +4,7 @@ package inox
 package evaluators
 
 object optIgnoreContracts extends FlagOptionDef("ignore-contracts", false)
+object optCrashOnChoose extends FlagOptionDef("crash-on-choose", false)
```
samarion: Nit: eval-quantifiers?

src/main/scala/inox/ast/SymbolOps.scala
```diff
@@ -115,42 +116,30 @@ trait SymbolOps { self: TypeOps =>
...
-   * @see [[canBeForced]]
-   */
-  def simplifyGround(expr: Expr)(implicit sem: symbols.Semantics, ctx: Context): Expr = {
-    val evaluator = sem.getEvaluator
+    * by evaluating them using [[inox.evaluators.Evaluator]].
+    * If `force` is omitted, the given expression will only be evaluated if it is
+    * ground, pure, and does not contain choose or quantifiers.
+    */
+  def simplifyGround(expr: Expr, force: Boolean = false)
+                    (implicit sem: symbols.Semantics, ctx: Context, opts: PurityOptions): Result[Expr] = {
```
samarion: Ah I didn't realize you needed the context for getting the evaluator. You can get the PurityOptions from the context as `PurityOptions(ctx)`.

src/main/scala/inox/solvers/princess/PrincessSolver.scala
```diff
@@ -36,6 +36,7 @@ trait PrincessSolver extends AbstractUnrollingSolver { self =>
   object templates extends {
     val program: targetProgram.type = targetProgram
     val context = self.context
+    protected implicit val semantics: targetProgram.Semantics = self.targetSemantics
```
samarion: Nit: no need to make the override val implicit here.
Nor protected probably.

src/main/scala/inox/evaluators/Evaluator.scala
romac: Better name indeed! Thanks. Should I expose this option in the CLI, or do we keep only for internal usage?

src/main/scala/inox/solvers/princess/PrincessSolver.scala
romac: Indeed, fixed!

src/main/scala/inox/evaluators/Evaluator.scala
samarion: As you prefer.

romac:
Just fixed the bug in SMTLIBTarget which was making the unit tests fail.

romac:
It also looks like [`src/it/resources/regression/tip/UNKNOWN/PigeonHole.tip`](https://github.com/epfl-lara/inox/blob/e01a0a5ff2e423519e0d073f308d378779f6fac5/src/it/resources/regression/tip/UNKNOWN/PigeonHole.tip) now successfully verifies. If that is right, should I move it into `src/it/resources/regression/tip/UNSAT`?

samarion:
It seems like the `AbstractRefinementMap` tip test times out when assumeChecked mode is activated. Note that this mode basically makes everything pure. Could you post some debug of the input/outputs of the ground simplifier with and without assumeChecked? (you can run them from the command line)

samarion:
That is very strange. It would seem to me that the ground simplifier should have no effect on that program...

romac:
The "choose in parametric function" testcase in `ChooseSuite` also fails with [the following log](https://gist.github.com/romac/e2d3d803ce50946f16db3d70ab377078), but I am not sure why.

I get the same errors even after applying:

```diff
diff --git a/src/main/scala/inox/ast/SymbolOps.scala b/src/main/scala/inox/ast/SymbolOps.scala
index 5061c41c..6f388d17 100644
--- a/src/main/scala/inox/ast/SymbolOps.scala
+++ b/src/main/scala/inox/ast/SymbolOps.scala
@@ -127,6 +127,14 @@ trait SymbolOps { self: TypeOps =>
     val evalCtx = ctx.withOpts(evaluators.optEvalQuantifiers(false))
     val evaluator = sem.getEvaluator(ctx)

+    def containsQuantifiers(expr: Expr): Boolean = {
+      exists {
+        case c: Choose => true
+        case f: Forall => true
+        case _ => false
+      } (expr)
+    }
+
     def evalChildren(e: Expr): Expr = e match {
       case Operator(es, recons) => recons(es.map(rec))
     }
@@ -134,13 +142,16 @@ trait SymbolOps { self: TypeOps =>
     def rec(e: Expr): Expr = e match {
       case e if isValue(e) =>
         e
-      case e if isPure(e) && isGround(e) =>
+      case e if isPure(e) && isGround(e) && !containsQuantifiers(e) =>
         evaluator.eval(e).result.getOrElse(evalChildren(e))
       case e =>
         evalChildren(e)
     }

-    rec(expr)
+    println("before: " + expr)
+    val res = rec(expr)
+    println("after: " + res)
+    res
   }

   private val typedIds: MutableMap[Type, List[Identifier]] =
```

romac:
> It seems like the AbstractRefinementMap tip test times out when assumeChecked mode is activated. Note that this mode basically makes everything pure. Could you post some debug of the input/outputs of the ground simplifier with and without assumeChecked? (you can run them from the command line)

Indeed, the error seems to stem from the combination of `--check-models` and `--assume-check`. Here are the results: https://gist.github.com/romac/69f63482af037f39154c8a979b508af8

romac:
> That is very strange. It would seem to me that the ground simplifier should have no effect on that program...

Here's the output for `UNSAT/PigeonHole.tip`:

```
before: h1 ≠ h2 && h1 ≠ h3 && h1 ≠ h4 && h2 ≠ h3 && h2 ≠ h4 && h3 ≠ h4 && p1 ≠ p2 && p1 ≠ p3 && p1 ≠ p4 && p1 ≠ p5 && p2 ≠ p3 && p2 ≠ p4 && p2 ≠ p5 && p3 ≠ p4 && p3 ≠ p5 && p4 ≠ p5 && ∀p: hole. (p == h1 || p == h2 || p == h3 || p == h4)
before: true
after: true
before: true
after: true
after: false
[  Info  ]  => UNSAT
```

It looks like `simplifyGround` evaluates this first expression to `false`, before reaching evaluation of the quantifier. After forcing it to evaluate each conjunct separately, I get:

```
before: h1 ≠ h2 && h1 ≠ h3 && h1 ≠ h4 && h2 ≠ h3 && h2 ≠ h4 && h3 ≠ h4 && p1 ≠ p2 && p1 ≠ p3 && p1 ≠ p4 && p1 ≠ p5 && p2 ≠ p3 && p2 ≠ p4 && p2 ≠ p5 && p3 ≠ p4 && p3 ≠ p5 && p4 ≠ p5 && ∀p: hole. (p == h1 || p == h2 || p == h3 || p == h4)
before: true
after: true
before: true
after: true
before: true
after: true
before: true
after: true
before: true
after: true
before: true
after: true
before: true
after: true
before: true
after: true
before: true
after: true
after: false && false && false && false && false && false && false && false && false && false && false && false && false && false && false && false && ∀p: hole. (p == hole(0) || p == hole(0) || p == hole(0) || p == hole(0))
[  Info  ]  => UNSAT
```

Sounds like we can indeed move it into `UNSAT`, doesn't it?

src/main/scala/inox/ast/SymbolOps.scala
```diff
@@ -115,6 +116,50 @@ trait SymbolOps { self: TypeOps =>
...
+    }
+
+    def evalChildren(e: Expr): Expr = e match {
+      case Operator(es, recons) => recons(es.map(rec))
+    }
+
+    def rec(e: Expr): Expr = e match {
+      case e if isValue(e) =>
+        e
+      case e if force || isPure(e) && isGround(e) =>
```
samarion: Doesn't `||` bind weaker than `&&` in Scala? Your expression needs to _at least_ be ground to be evaluatable.
Also, I'm not sure I mentioned this or not but I still believe `force` should be handled only by Stainless. I see two options for force:
1. As you implemented it, force everything _under_ a function with the `@force` annotation. This basically means calling `simplifyGround` with `force = true` on the body of that function.
2. Evaluate all ground calls to functions that _have_ the `@force` annotation (should be implemented entirely outside of Inox).

Maybe @jad-hamza has some insights about which one would be most useful for proofs.

samarion:
About pigeon hole, each of the h_i and p_i terms should be function invocations where the function's body is a choose, so I would expect any call to `simplifyGround` to crash on the first conjunct (and certainly not output false!).

jad-hamza:
We could have both :) e.g. `@forceInBody` and `@forceCallsTo`. But can't we have the annotation at a finer level (on expressions)? Also, with the approach `@forceCallsTo`, will we reduce calls to the functions that appear under other lambdas (possibly making invalid VCs become valid by adding the annotation)?

samarion:
But the `@force` name is so much nicer... :)
We could have them at a finer level but I'm not sure they'll be useful since we'll be running the evaluation anyway in Inox. The only extra thing `@force` brings you is evaluation of contracts (and maybe non-pure expressions?). Once `assumeChecked` becomes the default in Stainless, pretty much all expressions will be pure.

If we ensure `@force` only evaluates pure expressions, validity of VCs won't be influenced by the annotation. Either the call was pure and was lifted for lambda equality, or it's impure and `@force` is ignored.

romac:
> But the @force name is so much nicer... :)
+1 ;)

> Once assumeChecked becomes the default in Stainless, pretty much all expressions will be pure.
What needs to be done to get that in? Would be tremendously useful now that we have the ground simplifier. Then yeah @force might only be useful to force evaluation of known non total expressions. And in that case, it might be ok to have it do so, even if it influences VC validation, as long as we maybe raise a warning.

samarion:
`assumeChecked` breaks a couple benchmarks that were relying on some contracts that get simplified. There are some in imperative and the `IntSet` benchmarks fail. I'm not too happy about breaking those since Viktor presented them at ScalaDays... However, there's no way to have both `assumeChecked` simplifications and those benchmarks passing so it's just a question of me accepting this break.

I think basically everything is total with `assumeChecked` IIRC. We assume all potentially impure code will be checked so everything can be assumed pure. So there aren't really any non total expressions. Just quantifiers that won't get simplified by `simplifyGround`.

jad-hamza:
> the IntSet benchmarks fail

I remember having the same problem in my control branch. The problem is that the variable `res` that holds the body of the function gets simplified away, right? Then we can make the tests pass by preventing the variable `res` from getting removed. 

I think we should make these tests pass because they're a valid way of writing the program and specification.

src/main/scala/inox/ast/SymbolOps.scala
romac: > Doesn't || bind weaker than && in Scala? Your expression needs to at least be ground to be evaluatable.

It does indeed but my thinking that it might sometimes be possible to evaluate a non-ground expression if eg. the free variable is located in a branch that is never entered. It's definitely an edge-case and I don't mind only evaluating ground expressions only.

> Also, I'm not sure I mentioned this or not but I still believe force should be handled only by Stainless.

It's likely, in which case I must have misunderstood you.

> I see two options for force:

> 1. As you implemented it, force everything under a function with the @force annotation. This basically means calling simplifyGround with force = true on the body of that function.
> 2. Evaluate all ground calls to functions that have the @force annotation (should be implemented entirely outside of Inox).
> Maybe @jad-hamza has some insights about which one would be most useful for proofs.

Ah good idea, the second one actually seems more useful to me. What do you think @jad-hamza ?

jad-hamza: Sounds good to me! I guess the second option gives you finer grained control, because you could always wrap the only expressions you want to force in a fresh function, annotated by @force (e.g. in a function named "force")

samarion: Ah I hadn't though of the non-ground branch. I still wouldn't try to run an evaluator on a non-ground expression (especially the `CodeGen` one). You can always run this within a fixpoint if you want with `simplifyByConstructor` which should simplify all your branching trees.

samarion:
One thing you still need to handle is avoiding ground evaluation under lambdas unless they are _always_ pure since this can affect the equality semantics. See `SymbolOps.isAlwaysPure`. You should keep a boolean variable during your recursive traversal which tells you whether you're in a lambda or not.

This seems disjoint from the type errors you're seeing in the choose and flatMap benchmarks with Princess. I don't really know what's going on there, I'll need to take a closer look (or you can post a bit more debug maybe).

samarion:
I fixed the remaining issues (see https://github.com/epfl-lara/inox/commit/7286357e6981754fee4a22b356ede2ab66666c47) and merged.

I kept the force option, but I'm not sure what it's intended goal is exactly. If you call `simplifyGround` with `PurityOptions.AssumeChecked`, you'll basically get the same behavior as with `force = true`. I would suggest either removing the `force` option or changing it to allow quantifier and choose evaluation when `force = true` (so it's at least useful for something).

Closing in the meantime since it's been merged, please implement the option you decide on in a new PR. Thanks!

romac:
@samarion Ok, thanks! I will remove the `force` option and update the stainless PR to evaluate ground calls to functions annotated with `@force`, as discussed in that PR.

PR #49, opened 2017-11-16
Update Timer API with runAndGetTime

mantognini:
The usage of this function is quite limited ATM but since I needed it in a third place I thought it was time to implement it. See the related PR in stainless.


samarion:
LGTM

PR #51, opened 2017-11-18
Fix ASCII table header & remove dead code

mantognini:
When the title contains some ASCII color, the padding was wrong.


samarion:
LGTM, you can rebase and merge. I'll publish a version once this is in so we can update Stainless with both our changes.

PR #53, opened 2017-11-27
Remove `force` argument on `simplifyGround`

romac:


romac:
@samarion Thanks! Can you pleasd ping me once this version is published on Maven, so that I can update https://github.com/epfl-lara/stainless/pull/108?

samarion:
Hmm, I think you should implement https://github.com/epfl-lara/stainless/pull/108 without relying on `simplifyGround` so you can enable evaluation of quantifiers and chooses. Since you only want to evaluate single expressions and not recurse into children, you can ignore the `isPure` condition and use something as simple as:
```scala
val evaluator = semantics.getEvaluator
private def force(fi: FunctionInvocation): Expr = {
  reporter.debug(s"Forcing invocation of `${fi.id}` @ ${fi.getPos}...")
  reporter.debug(s" - Before: $fi...")
  val res = evaluator.eval(fi) match {
    case Success(e) => exprOps.freshenLocals(e)
    case _ => // report errors
  }
  reporter.debug(s" - After: $res")
  res
}
```

PR #55, opened 2017-12-20
Print tip filename with result

mantognini:
I was going through my local branches and though this could be useful to others if you happen to run inox with several input files. :)


mantognini:
In that case, I left it as is. Feel free to update it later on. ;-)

PR #56, opened 2017-12-20
Ensure ASCIIStringEncoder does not attempt to decode strings of length < 2

romac:
Previously, when decoder would attempt to read the first two bytes
of any non-empty string it was given, which obviously fails when the
string has length 1. This commit fixes that by directly returning
any string of length < 2.

I ran into this bug while running the termination checker over this benchmark:

```scala
import stainless.lang._
import stainless.annotation._
import stainless.collection._

object kv {

  sealed abstract class Label
  object Label {
    case class Get(key: String) extends Label
    case class Put(key: String, value: String) extends Label
  }

  sealed abstract class Op
  case class Pure(value: Option[String]) extends Op
  case class Get(key: String, next: Option[String] => Op) extends Op
  case class Put(key: String, value: String, next: () => Op) extends Op

  def get(key: String)(next: Option[String] => Op): Op = Get(key, next)
  def put(key: String, value: String)(next: () => Op): Op = Put(key, value, next)
  def pure(value: Option[String]): Op = Pure(value)

  val program =
    put("foo", "bar") { () =>
      get("foo") { value =>
        pure(value)
      }
    }

  def interpret(op: Op)(kv: Map[String, String], trace: List[Label], fuel: Int): (Option[String], List[Label]) = {
    require(fuel > 0)
    decreases(fuel)

    op match {
      case Get(key, next) =>
        interpret(next(kv get key))(kv, Label.Get(key) :: trace, fuel - 1)

      case Put(key, value, next) =>
        interpret(next())(kv.updated(key, value), Label.Put(key, value) :: trace, fuel - 1)

      case Pure(value) =>
        (value, trace)
    }
  }

  def test(map: Map[String, String], trace: List[Label]) = {
    val (res, resTrace) = interpret(program)(map, trace, 100)
    res == Some("bar") &&
    resTrace.take(2).reverse == List(Label.Put("foo", "bar"), Label.Get("foo"))
  } holds

}
```

Should I submit it to stainless, or should I rather add the TIP output from the termination checker to inox' test suite, or both?

### DIFF
--- a/src/main/scala/inox/solvers/theories/ASCIIStringEncoder.scala
+++ b/src/main/scala/inox/solvers/theories/ASCIIStringEncoder.scala
@@ -83,7 +83,7 @@ trait ASCIIStringEncoder extends SimpleEncoder {
 
     override def transform(e: Expr): Expr = e match {
       case ADT(String, Seq(StringLiteral(s))) =>
-        def unescape(s: String): String = if (s.isEmpty) s else {
+        def unescape(s: String): String = if (s.length < 2) s else {
           val (b1, s2) = decodeFirstByte(s)
           val (b2, s3) = decodeFirstByte(s2)
           val h: String = if (0 <= b1 && b1 <= 127 && b1 == b2) {


samarion:
The encoder should always output two characters for every input character, see 
https://github.com/epfl-lara/inox/blob/1ef50958bee5bf7762cad64a7bfe71b73d726302/src/main/scala/inox/solvers/theories/ASCIIStringEncoder.scala#L65-L69
so this must be symptomatic of some other bug.

I'll try to take a look sometime during the holidays. Or you can print the encoded strings and maybe I can debug the issue from there.

samarion:
I can't reproduce your error unfortunately. It's possible we get this issue for parts of the model returned by the solver that don't need to conform to the ADT invariant. I don't know how best to handle this case though...

romac:
I too am now unable to reproduce it with latest `master` of both Stainless and Inox. Will close for now, and see if this error pops up again or if I can try to reproduce it once I can dedicate time to it.

PR #62, opened 2018-04-05
Store ill-typed expression within TypeErrorException

romac:
It would nice to be able to access the expression that caused a TypeErrorException to be throwned, eg. to show the typing derivation to the user or just for debugging purposes.


samarion:
Merged, thanks!

PR #63, opened 2018-04-13
Allow annotations on ADT constructors

romac:


romac:
Closing after further discussion with @samarion.

PR #64, opened 2018-04-13
Replace sort-level flag for invariants by a function-level flag

romac:


samarion:
I'm not sure we want this within Inox. The `IsInvariantOf` annotation for `ADTConstructor` doesn't make any sense here and makes the API more complex. I think having both the `HasADTInvariant` and `IsInvariantOf` flags side-by-side in Stainless is fine and keeps a cleaner API in Inox.

romac:
Makes sense!

PR #71, opened 2018-05-19
Cross-compilation support for Scala 3 (WIP)

gsps:
This is a WIP branch to track the progress on making inox cross-compile on Scala 2 and Scala 3. So far, we can compile and manually run inox on test cases using the CLI and Z3.

The remaining tasks roughly comprise:

- [ ] Fix the sbt build to support cross-compilation out of the box (currently fixed to dotc).
- [ ] Reactivate `TipDebugger` and the core methods of `TimeoutSolver` (blocked by dotc bug around `abstract override`).
- [ ] Reactivate `Variable#copy` (blocked by dotc bug that raises denotation merge error). The synthetic copy method is still there, but doesn't preserve positions now.
- [ ] Reactivate `BagEncoder` and `SetEncoder` (blocked by dotc bug on vararg applications). As a result we can also fix up CVC4 and reactivate its code.
- [ ] Fix up Princess.
- [ ] Ideally, make some of the test suite work on dotc (scalatest is currently unavailable, since it relies on macros).
- [ ] Reactivate `@inline`s, dotc's inliner just causes too much trouble at the moment.

This PR includes workarounds for several other dotc issues and also some fixes for things that are presumably (non-serious) bugs in inox.

We track the related dotc issues here:
- [ ] https://github.com/lampepfl/dotty/issues/4520 Surprising synthesized `_$n` selectors on case classes
- [x] https://github.com/lampepfl/dotty/issues/4558 Missing adaptation from by-name to result type
- [ ] https://github.com/lampepfl/dotty/issues/4559 Unintended mixin initializer for overriden lazy val
- [ ] https://github.com/lampepfl/dotty/issues/4563 Erasure tries to box call to label def
- [ ] https://github.com/lampepfl/dotty/issues/4564 Overriding `copy` of case class results in unhelpful error messages
- [ ] #??? https://github.com/epfl-lara/inox/pull/71/commits/8f50d7bf18b4a116b9bce1e9e6023c92f3c04b2a
- [ ] #??? https://github.com/epfl-lara/inox/pull/71/commits/96ccf5cfdba111f385e63bf60c105be6549ca0a3
- [ ] #??? https://github.com/epfl-lara/inox/pull/71/commits/210295d58bfcb953ceb27175a156d16f4b3420ce
- [ ] #??? https://github.com/epfl-lara/inox/pull/71/commits/d8ca991c541bc6ed02a79a59caff01aa2e444637
- [ ] #??? https://github.com/epfl-lara/inox/pull/71/commits/8450be02f8ef5b9311bd25d3b4b8f930e1d16feb
- [ ] #??? https://github.com/epfl-lara/inox/pull/71/commits/ea03b0679145b499c9558150d55a39d532112948

### DIFF
--- a/src/main/scala/inox/solvers/theories/ASCIIStringEncoder.scala
+++ b/src/main/scala/inox/solvers/theories/ASCIIStringEncoder.scala
@@ -108,7 +108,8 @@ trait ASCIIStringEncoder extends SimpleEncoder {
 }
 
 object ASCIIStringEncoder {
-  def apply(p: Program): ASCIIStringEncoder { val sourceProgram: p.type } = new {
-    val sourceProgram: p.type = p
-  } with ASCIIStringEncoder
+  def apply(p: Program): ASCIIStringEncoder { val sourceProgram: p.type } = {
+    class ASCIIStringEncoderBase(val sourceProgram: p.type) extends ASCIIStringEncoder
+    new ASCIIStringEncoderBase(p)
+  }
 }
--- a/src/main/scala/inox/solvers/theories/ASCIIStringEncoder.scala
+++ b/src/main/scala/inox/solvers/theories/ASCIIStringEncoder.scala
@@ -43,11 +43,14 @@ trait ASCIIStringEncoder extends SimpleEncoder {
   protected object encoder extends SelfTreeTransformer {
     override def transform(e: Expr): Expr = e match {
       case StringLiteral(v) =>
-        stringCons(StringLiteral(v.flatMap(c => c.toString.getBytes.toSeq match {
-          case Seq(b) if 32 <= b && b <= 127 => b.toChar.toString + b.toChar.toString
-          case Seq(b) => encodeByte(b) + encodeByte(b)
-          case Seq(b1, b2) => encodeByte(b1) + encodeByte(b2)
-        })))
+        // NOTE(gsps): [Bug] Workaround for erasure/adaptation bug in Dotty
+        def decode(c: Char): String =
+          c.toString.getBytes.toSeq match {
+            case Seq(b) if 32 <= b && b <= 127 => b.toChar.toString + b.toChar.toString
+            case Seq(b) => encodeByte(b) + encodeByte(b)
+            case Seq(b1, b2) => encodeByte(b1) + encodeByte(b2)
+          }
+        stringCons(StringLiteral(v.flatMap(decode)))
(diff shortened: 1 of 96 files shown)


src/main/scala/inox/solvers/theories/ASCIIStringEncoder.scala
```diff
@@ -108,7 +108,8 @@ trait ASCIIStringEncoder extends SimpleEncoder {
 }
 
 object ASCIIStringEncoder {
-  def apply(p: Program): ASCIIStringEncoder { val sourceProgram: p.type } = new {
-    val sourceProgram: p.type = p
-  } with ASCIIStringEncoder
+  def apply(p: Program): ASCIIStringEncoder { val sourceProgram: p.type } = {
```
smarter: The local class shouldn't be needed, you should be able to directly write:
```scala
def apply(p: Program): ASCIIStringEncoder { val sourceProgram: p.type } = new ASCIIStringEncoder {
  val sourceProgram: p.type = p
}
````

samarion: The methods used at the beginning of the `ASCIIStringEncoder`
https://github.com/epfl-lara/inox/blob/d13789665dbc9b8ce7a32557d68460d11307fa67/src/main/scala/inox/solvers/theories/ASCIIStringEncoder.scala#L17-L28 like `mkSort`, `mkFunDef`, `T`, etc. depend on
https://github.com/epfl-lara/inox/blob/d13789665dbc9b8ce7a32557d68460d11307fa67/src/main/scala/inox/solvers/theories/TheoryEncoder.scala#L11 so we need the `sourceProgram` to be initialized before the `ASCIIStringEncoder` constructor gets called. We could maybe use lazy vals for all these fields but we had a bunch of deadlocking issues due to laziness in the project so we'd prefer not having to introduce new ones.

Trait parameters would be nicer but we'd like to keep source compatibility with 2.11 for now.

romac:
Dotty 0.15.0-RC1 now fully bootstraps (https://github.com/lampepfl/dotty/pull/5923), so it’s perhaps time to give another go at this PR :)

src/main/scala/inox/solvers/theories/ASCIIStringEncoder.scala
romac: Do you still want to maintain source compatibility with 2.11 now that 2.13 is around the corner? Supporting the last two Scala version + Dotty sounds like a good deal even if that means dropping 2.11 no?

PR #75, opened 2018-08-25
Add filtering capability for Symbols based on names

jad-hamza:
Can be used to filter ADTs and functions from Symbols based on their names


jad-hamza:
I use it here for debugging purposes: https://github.com/epfl-lara/stainless/pull/308/files, to only display functions and classes that the user wants to see (specified by `--debug-objects`). 

I could also write `filterObjects` externally instead? No need to modify Inox this way.

samarion:
It's probably easiest if we only modify Stainless to avoid having to re-publish Inox.
But in general, it's probably better to filter during printing instead of building an invalid Symbol table, for example when printing with attached types (they won't be computable anymore if you remove dependencies from the symbol table).

jad-hamza:
Yes you're right, I'll do the change.

PR #76, opened 2018-08-30
Adding partial support for unsigned bitvector operations

jad-hamza:


jad-hamza:
Actually Inox tests are passing (the error I got locally was due to https://github.com/epfl-lara/inox/issues/77), I'll now work on the Stainless ones.

PR #79, opened 2018-09-04
Negation of environment using not(and(_)) instead of map

jad-hamza:
I'm not sure if that's the right way to use `Env`, but this does seem to fix https://github.com/epfl-lara/stainless/issues/320. I hope it doesn't break anything


jad-hamza:
@romac pointed me to this one as well: https://github.com/epfl-lara/inox/blob/99de42efea6c0065b3ef6ea5f75c98b0c9ff03b4/src/main/scala/inox/transformers/SimplifierWithCNFPath.scala#L175

PR #84, opened 2018-10-29
Complete rewrite of the parser

redelmann:


samarion:
Hey @redelmann, is there any progress on this? Is the rewrite in a mergeable state?

redelmann:
Hey @samarion, I don't think it is. I will look at this when I have time!

PR #92, opened 2019-03-06
Respect PrinterOptions when throwing a TypeErrorException

romac:
Make it easier to debug some type errors, eg. when we get `expected 'Type', found 'Type'`.


samarion:
This `Lazy1` cache seems dangerous. I'm not convinced we won't have cache corruptions due to concurrency here.

WDYT about building an interface for Inox exceptions that has an `asString` method which takes a context? This could clean up the exception messages in Inox a bit.

romac:
> WDYT about building an interface for Inox exceptions that has an asString method which takes a context? This could clean up the exception messages in Inox a bit.

Sounds good :)

PR #98, opened 2019-05-27
Parsing and extraction for lambdas in Z3's (4.8.4) model output

jad-hamza:
Z3 4.8.4 returns lambdas which can be parsed using `smtlib.extensions.tip`. These can then be extracted into maps, sets and bags similarly to the `as-array` cases.

### DIFF
--- a/src/main/scala/inox/solvers/smtlib/Z3Target.scala
+++ b/src/main/scala/inox/solvers/smtlib/Z3Target.scala
@@ -91,7 +91,7 @@ trait Z3Target extends SMTLIBTarget with SMTLIBDebugger {
   protected val extSym = SSymbol("_")
 
   protected lazy val setSort: SSymbol = {
-    val s = SSymbol("Set")
+    val s = SSymbol("InoxSet")
     val t = SSymbol("T")
 
     val arraySort = Sort(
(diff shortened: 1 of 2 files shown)


vkuncak:
This is important since we want to move beyond buggy previous versions of Z3

samarion:
I'm not convinced that Z3 actually removes bugs when increasing the version...

@jad-hamza: why aren't we using Z3's sets in version >= 4.8.4 if they now exist?

vkuncak:
4.7.1 has a particular bug with modulo that we really do not want to have and it seems gone in our test cases in later version. (If we have some other specific bugs that we know about, we can "watch" for those as well.) For sets, I guess we should first decide in Stainless if this refers to finite or potentially infinite sets.

jad-hamza:
I'm not sure sets are even in 4.8.4 (only in master), for now I only changed the name to avoid name clashes

jad-hamza:
Of course it'd be good to move to their sets eventually if they are compatible with what Inox is expecting.

samarion:
Published with version 1.1.0-329-g3c23a86

jad-hamza:
Nice, thanks!

PR #99, opened 2019-06-06
Check purity of tuples by checking that all components are pure

jad-hamza:
Fixes https://github.com/epfl-lara/stainless/issues/531


romac:
@samarion Do you think you can find time to review this in the next few days?

PR #102, opened 2019-06-29
Move implicit debug section higher

jad-hamza:
I sometimes have this error when compiling Inox from Stainless (this PR fixes it)

```
inox/src/main/scala/inox/solvers/unrolling/Templates.scala:79:19: could not find implicit value for parameter section: inox.DebugSection
[error]     reporter.debug("Unrolling generation [" + currentGen + "]")
```


samarion:
Subsumed by https://github.com/epfl-lara/inox/pull/105

PR #103, opened 2019-06-30
Remove merge functions

jad-hamza:
Fix https://github.com/epfl-lara/stainless/issues/554.

From quick tests, this didn't appear to affect performance for Stainless and Bolts tests.


jad-hamza:
More tests on Stainless using `it:testOnly *VerificationSuite*`: on this branch 501s, and on master with `mergeFunctions` still there: 467s. We'd have to repeat the experiments to see if it's not variance, but it might be worth it to try to fix it.

jad-hamza:
Instead of removing `mergeFunctions`, I've tried reordering the simplifications
(a valid workaround for https://github.com/epfl-lara/stainless/issues/554),
by replacing every `compose` with `andThen` in:

https://github.com/epfl-lara/inox/blob/a6cbf8e2488fd2d3aa941963fca65eb5d370c698/src/main/scala/inox/ast/SymbolOps.scala#L1204-L1220

I ran 4 tests on the [LeaderElection example](https://github.com/epfl-lara/stainless-actors) from the `stainless-scalac` command-line (with `smt-z3` 4.8.5 and no cache):

* only `compose`: 9m41s
* only `andThen`: 10m25s
* only `compose`: 9m19s
* only `andThen`: 11m56s

The `compose` version seems to be faster so let's not reorder for the moment.

PR #119, opened 2020-10-28
Fix the extraction of finite maps from Z3

brunnerant:
This PR fixes a special case where finite maps are not correctly extracted. This happens for the special case where the value type of the map is `BooleanType`. Z3 sometimes represents those in that form: `(lambda ((x KeyType)) (= x someKey))`. That should be extracted as a finite map where `someKey` maps to `true` and the rest to `false`.

That case is currently not handled and I encountered a case where the output from stainless was misleading because of that extraction problem.

### DIFF
--- a/src/main/scala/inox/solvers/smtlib/Z3Target.scala
+++ b/src/main/scala/inox/solvers/smtlib/Z3Target.scala
@@ -148,6 +148,8 @@ trait Z3Target extends SMTLIBTarget with SMTLIBDebugger {
         val Some(Lambda(Seq(arg), body)) = context.getFunction(k, FunctionType(Seq(keyType), valueType))
 
         def extractCases(e: Expr): FiniteMap = e match {
+          case Equals(argV, k) if valueType == BooleanType() && argV == arg.toVariable =>
+            FiniteMap(Seq((k, BooleanLiteral(true))), BooleanLiteral(false), keyType, valueType)
           case IfExpr(Equals(argV, k), v, e) if argV == arg.toVariable =>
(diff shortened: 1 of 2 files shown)


brunnerant:
Are there some test cases that treat the extraction from SMTLIB ? If that's the case, I am happy to add some cases for what I have added.

gsps:
I think you could add a case to [`SemanticsSuite`](https://github.com/epfl-lara/inox/blob/master/src/test/scala/inox/solvers/SemanticsSuite.scala). The tests there take an Inox tree `t`, lower it to a simple query `v = t` and compare the model found for `v` against an expected Inox tree.

brunnerant:
The thing is, I want to test that certain representations of sets in SMTLIB are extracted correctly, so I am not sure how I can leverage that testing infrastructure. I have no guarantee that the model that SMTLIB returns for `v` has the format that I want to test...

gsps:
That's true, but perhaps you could do something more like an integration test that asserts some operations on `Set`s yield `FiniteSet`s we can extract from models. It might be that the new pattern only arises for our `MapMerge` operation, in which case we can't effectively test this on Inox's master, but you could still add the test to your branch.

brunnerant:
Yes that's probably something that I have to dig into. I will first try to understand how this representation of sets arises from SMTLIB.

brunnerant:
I found a way to test the particular feature that I added. Somehow, when encoding a singleton set inside a map, Z3 uses the form that I mentioned above `(lamdba ((x Int)) (= x 0))`.

samarion:
LGTM

PR #121, opened 2020-11-25
Add MapMerge construct

gsps:
This PR adds a new tree `MapMerge(mask, map1, map2)` which takes a set `mask : Set[K]` and two maps `map1, map2 : Map[K, V]`. It produces a map combining `map1` and `map2`, picking values from `map1` whenever `k elemOf mask`.

This is encoded efficiently in the extended array theory of Z3. For other solvers we encode `MapMerge(mask, map1, map2)` as `Choose(m:Map[K,V] => forall x:K. m(k) = if (mask(k)) map1(k) else map2(k))`. In practice, CVC4 and Princess seem to simply return `Unknown` in the presence of `MapMerge`.

I added a few (very basic) tests for `nativez3`, `unrollz3` and `smt-z3`.

### DIFF
--- a/src/it/scala/inox/solvers/unrolling/MapMergeSuite.scala
+++ b/src/it/scala/inox/solvers/unrolling/MapMergeSuite.scala
@@ -1,0 +1,66 @@
+/* Copyright 2009-2020 EPFL, Lausanne */
+
+package inox
+package solvers
+package unrolling
+
+class MapMergeSuite extends SolvingTestSuite with DatastructureUtils {
+  import trees._
+  import dsl._
+
+  override def configurations = for {
+    solverName   <- Seq("nativez3", "unrollz3", "smt-z3")
+    feelingLucky <- Seq(false, true)
+  } yield Seq(
+    optSelectedSolvers(Set(solverName)),
+    optCheckModels(true),
+    optFeelingLucky(feelingLucky),
+    optTimeout(300),
+    ast.optPrintUniqueIds(true)
+  )
+
+  override def optionsString(options: Options): String = {
+    "solvr=" + options.findOptionOrDefault(optSelectedSolvers).head + " " +
+    "lucky=" + options.findOptionOrDefault(optFeelingLucky)
+  }
+
+  val keyT = TypeParameter.fresh("K")
+  val valueT = TypeParameter.fresh("V")
+
+  val mask = ("mask" :: SetType(keyT)).toVariable
+  val map1 = ("map1" :: MapType(keyT, valueT)).toVariable
+  val map2 = ("map2" :: MapType(keyT, valueT)).toVariable
+  val merged = MapMerge(mask, map1, map2)
+  val dummyValue = ("dummyValue" :: valueT).toVariable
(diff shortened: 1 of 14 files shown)


src/main/scala/inox/solvers/theories/MapMergeEncoder.scala
```diff
@@ -0,0 +1,66 @@
...
+      case e: MapMerge => transform(MapMergeEncoded(e))
+      case _ => super.transform(e)
+    }
+  }
+
+  protected object decoder extends SelfTreeTransformer {
+    import targetProgram._
+
+    override def transform(e: Expr): Expr = e match {
+      case MapMergeEncoded(mask, map1, map2) =>
```
gsps, line 55: @samarion Could we actually ever run into the encoded form of `MapMerge` in a model?

samarion: Yes, it could appear when decoding a model which contains an input lambda containing a `MapMerge`. A bit of an edge case, but still best to cover it.

src/it/scala/inox/solvers/unrolling/MapMergeSuite.scala
```diff
@@ -0,0 +1,66 @@
...
+package inox
+package solvers
+package unrolling
+
+class MapMergeSuite extends SolvingTestSuite with DatastructureUtils {
+  import trees._
+  import dsl._
+
+  override def configurations = for {
+    solverName   <- Seq("nativez3", "unrollz3", "smt-z3")
```
samarion, line 12: I'm surprised none of the tests in this file work with the encoding. Did you take a look at why they were failing?

src/main/scala/inox/solvers/theories/MapMergeEncoder.scala
```diff
@@ -0,0 +1,66 @@
+/* Copyright 2009-2020 EPFL, Lausanne */
+
+package inox
+package solvers
+package theories
+
+trait MapMergeEncoder extends SimpleEncoder {
+  import trees._
+
+  object MapMergeEncoded {
```
samarion, line 10: One small improvement in the encoder would be to hide the choose below a `FunDef`. This would help the solvers prove programs involving `MapMerge` and also make the decoding step easier.

src/it/scala/inox/solvers/unrolling/MapMergeSuite.scala
gsps: They all return `Unknown` from what I remember (both in CVC4 and Princess).

gsps:
Moved to wrapping the encoding for `MapMerge` in `MapMergeEncoder` within a dedicated function. Didn't help with CVC4/Princess, unfortunately (at least on the few test cases in `MapMergeSuite`).

samarion:
Yeah I wouldn't expect the change to help for counter-example finding. I'm a bit surprised it doesn't work for such simple cases but I assume the quantifier instantiation procedure doesn't like some part of the choose. I don't think it matters too much.

PR #123, opened 2021-01-06
Avoid simplification during transformation

jad-hamza:
Fix https://github.com/epfl-lara/stainless/issues/880

### DIFF
--- a/src/main/scala/inox/Reporter.scala
+++ b/src/main/scala/inox/Reporter.scala
@@ -61,11 +61,15 @@ abstract class Reporter(val debugSections: Set[DebugSection]) {
     if (!cond) internalError(pos,msg)
   }
 
+  def reset() = {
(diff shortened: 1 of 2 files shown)


src/main/scala/inox/Reporter.scala
```diff
@@ -61,11 +64,22 @@ abstract class Reporter(val debugSections: Set[DebugSection]) {
     if (!cond) internalError(pos,msg)
   }
 
+  def reset() = {
+    _errorCount = 0
+    _warningCount = 0
```
samarion, line 69: `_fatalCount` be reset here as well.

jad-hamza: Oops, thanks!

jad-hamza:
Thanks!

PR #125, opened 2021-02-10
Change set and bag types encoding to account for emptiness (thanks @samarion)

jad-hamza:


samarion:
Thanks for adding the fix!

Could you maybe also add a test suite similar to the [MapMergeSuite](https://github.com/epfl-lara/inox/blob/master/src/it/scala/inox/solvers/unrolling/MapMergeSuite.scala) which checks that counterexamples to dependent types exist and are correct?

jad-hamza:
I added a test suite, but the new map tests fail: http://laraquad4.epfl.ch:9000/epfl-lara/inox/build/31/162
I probably messed things up when porting the code from `SetType`.
(I tried with the previous encoding for `MapType` and these tests went through)

samarion:
I think the test is wrong. The empty map is a valid value for the test "Maps with empty key type don't exist".
You can just assert `isSat` instead :)

Thanks for adding the test!

jad-hamza:
My bad, thanks!

PR #126, opened 2021-02-10
Fix inlinePosts

jad-hamza:
Fix https://github.com/epfl-lara/stainless/issues/568

> I'm not 100% sure how best to fix this... as a temporary solution, you could replace collect with transformWithPC and make sure the generated conjuncts are wrapped with the relevant path.

I used `collectWithPC`, is that ok? And I used `implies` to wrap the path, not sure if that's good.

### DIFF
--- a/src/main/scala/inox/ast/ExprOps.scala
+++ b/src/main/scala/inox/ast/ExprOps.scala
@@ -69,7 +69,7 @@ trait ExprOps extends GenTreeOps {
     * Note that we don't freshen choose ids as these are considered global
     * and used to lookup their images within models!
     */
-  object Freshener extends SelfTransformer {
+  class Freshener(freshenChooses: Boolean = false) extends SelfTransformer {
     type Env = Map[Identifier, Identifier]
 
     override def transform(e: Expr, env: Env): Expr = e match {
@@ -95,8 +95,12 @@ trait ExprOps extends GenTreeOps {
         }
         Forall(sparams, transform(b, senv)).copiedFrom(e)
 
-      case Choose(res, pred) =>
-        Choose(transform(res, env), transform(pred, env)).copiedFrom(e)
+      case Choose(vd, pred) if !freshenChooses =>
+        Choose(transform(vd, env), transform(pred, env)).copiedFrom(e)
+
+      case Choose(vd, pred) if freshenChooses =>
+        val freshVd = vd.freshen
+        Choose(transform(freshVd, env), transform(pred, env.updated(vd.id, freshVd.id))).copiedFrom(e)
 
       case _ =>
         super.transform(e, env)
@@ -127,8 +131,12 @@ trait ExprOps extends GenTreeOps {
     }
   }
 
-  def freshenLocals(expr: Expr): Expr = {
-    Freshener.transform(expr, Map[Identifier, Identifier]())
+  val freshenerNoChooses = new Freshener(false)
+  val freshenerWithChooses = new Freshener(true)
+
+  def freshenLocals(expr: Expr, freshenChooses: Boolean = false): Expr = {
+    if (freshenChooses) freshenerWithChooses.transform(expr, Map.empty[Identifier, Identifier])
+    else freshenerNoChooses.transform(expr, Map.empty[Identifier, Identifier])
   }
 
   /** Returns true if the expression contains a function call */
--- a/src/main/scala/inox/ast/ExprOps.scala
+++ b/src/main/scala/inox/ast/ExprOps.scala
@@ -41,7 +41,7 @@ trait ExprOps extends GenTreeOps {
   def replaceFromSymbols[V <: VariableSymbol](substs: Map[V, Expr], expr: Expr)(implicit ev: VariableConverter[V]): Expr = {
     new SelfTreeTransformer {
       override def transform(expr: Expr): Expr = expr match {
-        case v: Variable => substs.getOrElse(v.to[V], super.transform(v))
+        case v: Variable => freshenLocals(substs.getOrElse(v.to[V], super.transform(v)))
         case _ => super.transform(expr)
       }
     }.transform(expr)
@@ -70,21 +70,21 @@ trait ExprOps extends GenTreeOps {
     * and used to lookup their images within models!
     */
   def freshenLocals(expr: Expr, freshenChooses: Boolean = false): Expr = {
-    val subst: MutableMap[Variable, Variable] = MutableMap.empty
-    variablesOf(expr).foreach(v => subst(v) = v)
+    val subst: MutableMap[Identifier, Identifier] = MutableMap.empty
+    variablesOf(expr).foreach(v => subst(v.id) = v.id)
 
     new SelfTreeTransformer {
-      override def transform(vd: ValDef): ValDef = subst.getOrElse(vd.toVariable, {
-        val res = super.transform(vd).freshen.toVariable
-        subst(vd.toVariable) = res
+      override def transform(vd: ValDef): ValDef = {
+        val res = super.transform(vd).freshen
+        subst(vd.id) = res.id
         res
-      }).toVal
+      }
 
       override def transform(expr: Expr): Expr = expr match {
-        case v: Variable => transform(v.toVal).toVariable
+        case v: Variable => v.copy(id = subst(v.id)).copiedFrom(v)
         case Choose(res, pred) if !freshenChooses =>
           val newVd = super.transform(res)
-          subst(res.toVariable) = newVd.toVariable
+          subst(res.id) = newVd.id
           Choose(newVd, transform(pred)).copiedFrom(expr)
         case _ => super.transform(expr)
       }
--- a/src/main/scala/inox/ast/ExprOps.scala
+++ b/src/main/scala/inox/ast/ExprOps.scala
@@ -69,7 +69,7 @@ trait ExprOps extends GenTreeOps {
     * Note that we don't freshen choose ids as these are considered global
     * and used to lookup their images within models!
     */
-  class Freshener(freshenChooses: Boolean = false) extends SelfTransformer {
+  protected class Freshener(freshenChooses: Boolean) extends SelfTransformer {
     type Env = Map[Identifier, Identifier]
 
     override def transform(id: Identifier, env: Env): Identifier = {
@@ -96,9 +96,6 @@ trait ExprOps extends GenTreeOps {
             (sparams :+ transform(freshVd, env), env.updated(vd.id, freshVd.id))
         }
         Forall(sparams, transform(b, senv)).copiedFrom(e)
-
-      case Choose(vd, pred) if !freshenChooses =>
-        Choose(transform(vd, env), transform(pred, env)).copiedFrom(e)
 
       case Choose(vd, pred) if freshenChooses =>
         val freshVd = vd.freshen
@@ -133,12 +130,8 @@ trait ExprOps extends GenTreeOps {
     }
   }
 
-  val freshenerNoChooses = new Freshener(false)
-  val freshenerWithChooses = new Freshener(true)
-
   def freshenLocals(expr: Expr, freshenChooses: Boolean = false): Expr = {
-    if (freshenChooses) freshenerWithChooses.transform(expr, Map.empty[Identifier, Identifier])
-    else freshenerNoChooses.transform(expr, Map.empty[Identifier, Identifier])
+    new Freshener(freshenChooses).transform(expr, Map.empty[Identifier, Identifier])
   }
 
   /** Returns true if the expression contains a function call */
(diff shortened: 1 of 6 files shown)


samarion (approved):

jad-hamza:
> samarion approved these changes

larabot didn't approve :(
It's probably because `path implies _` ignores the `OpenBound`'s?

samarion:
`OpenBound`s were ignored before, so that shouldn't be the issue.

Maybe you need some freshening in the generated clause.

Also skipping inlinings when the path contains open bounds is probably a good idea.

jad-hamza:
I had a look at the new problem with `Lang.scala-0.tip` but that doesn't seem related to `inlinePosts`, there are variables which are not fresh (`x$0` appears several times) before entering the simplification functions. Should everything be fresh before entering `mergeFunctions`?

```scala
// before mergeFunctions
¬fun1!4$0[List!4$0[Int], Boolean](Lang!3$0[Int](fun1!4$0[List!4$0[Int], Boolean]((x$0: List!4$0[Int]) => ¬∀x$1: BigInt. ¬¬¬(((x$1 <= size!0$0[Int](x$0) && x$1 >= 0) && contains!1$0[Int](Lang!3$0[Int](fun1!4$0[List!4$0[Int], Boolean]((x$0: List!4$0[Int]) => x$0 == Cons!3$0[Int](3, Nil!3$0[Int]()), (x$0: List!4$0[Int]) => true)), take!0$0[Int](x$0, x$1))) && contains!1$0[Int](Lang!3$0[Int](fun1!4$0[List!4$0[Int], Boolean]((x$0: List!4$0[Int]) => x$0 == Cons!3$0[Int](5, Nil!3$0[Int]()), (x$0: List!4$0[Int]) => true)), drop!0$0[Int](x$0, x$1))), (x$0: List!4$0[Int]) => true)).f!45$0.pre!12$0, (x!93$0: List!4$0[Int]) => true).f!46$0(x!56$0)

// after mergeFunctions
¬fun1!4$0[List!4$0[Int], Boolean](Lang!3$0[Int](fun1!4$0[List!4$0[Int], Boolean]((x$0: List!4$0[Int]) => ¬∀x$1: BigInt. {
  val call$43: BigInt = size!0$0[Int](x$0)
  ¬¬¬(((x$1 <= call$43 && x$1 >= 0) && {
    val call$38: List!4$0[Int] = take!0$0[Int](x$0, x$1)
    val call$37: Boolean = contains!1$0[Int](Lang!3$0[Int](fun1!4$0[List!4$0[Int], Boolean]((x$0: List!4$0[Int]) => x$0 == Cons!3$0[Int](3, Nil!3$0[Int]()), (x$0: List!4$0[Int]) => true)), call$38)
    call$37
  }) && {
    val call$42: List!4$0[Int] = drop!0$0[Int](x$0, x$1)
    val call$41: Boolean = contains!1$0[Int](Lang!3$0[Int](fun1!4$0[List!4$0[Int], Boolean]((x$0: List!4$0[Int]) => x$0 == Cons!3$0[Int](5, Nil!3$0[Int]()), (x$0: List!4$0[Int]) => true)), call$42)
    call$41
  })
}, (x$0: List!4$0[Int]) => true)).f!45$0.pre!12$0, (x!93$0: List!4$0[Int]) => true).f!46$0(x!56$0)
```

```scala
// before simplifyForalls
¬fun1!4$0[List!4$0[Int], Boolean](Lang!3$0[Int](fun1!4$0[List!4$0[Int], Boolean]((x$0: List!4$0[Int]) => ¬∀x$1: BigInt. {
  val call$43: BigInt = size!0$0[Int](x$0)
  ¬¬¬(((x$1 <= call$43 && x$1 >= 0) && {
    val call$38: List!4$0[Int] = take!0$0[Int](x$0, x$1)
    val call$37: Boolean = contains!1$0[Int](Lang!3$0[Int](fun1!4$0[List!4$0[Int], Boolean]((x$0: List!4$0[Int]) => x$0 == Cons!3$0[Int](3, Nil!3$0[Int]()), (x$0: List!4$0[Int]) => true)), call$38)
    call$37
  }) && {
    val call$42: List!4$0[Int] = drop!0$0[Int](x$0, x$1)
    val call$41: Boolean = contains!1$0[Int](Lang!3$0[Int](fun1!4$0[List!4$0[Int], Boolean]((x$0: List!4$0[Int]) => x$0 == Cons!3$0[Int](5, Nil!3$0[Int]()), (x$0: List!4$0[Int]) => true)), call$42)
    call$41
  })
}, (x$0: List!4$0[Int]) => true)).f!45$0.pre!12$0, (x!93$0: List!4$0[Int]) => true).f!46$0(x!56$0)
```

```scala
[info]   java.lang.AssertionError: assertion failed: CANNOT ADD x$0: List!4$0[Int] to
[info] val call$43: BigInt = size!0$0[Int](x$0)
[info] x$1 <= call$43 && (x$1 >= 0 && {
[info]   val call$38: List!4$0[Int] = take!0$0[Int](x$0, x$1)
[info]   true
[info] })
```

jad-hamza:
Apparently, expressions don't have unique identifiers after the parser, for instance:
```scala
¬fun1!4$0[List!4$0[Int], Boolean](Lang!3$0[Int](fun1!4$0[List!4$0[Int], Boolean]((x$0: List!4$0[Int]) => x$0 == Cons!3$0[Int](5, Nil!3$0[Int]()), (x$0: List!4$0[Int]) => true)).f!45$0.pre!12$0, (x!93$0: List!4$0[Int]) => true).f!46$0(x!56$0)
```
Is that normal? (Also `freshenLocals` isn't designed to freshen those)

samarion:
Well, lambda normalization will normalize identifiers, which can create duplicates. Lambda normalization is used in some places which aren't ideal, e.g. the recursive evaluator so you could be seeing bad interactions there.

Which parser do you mean? The TIP parser freshens all variable names.

Also, what do you mean about `freshenLocals` not being designed for certain variables?

jad-hamza:
> Well, lambda normalization will normalize identifiers, which can create duplicates. Lambda normalization is used in some places which aren't ideal, e.g. the recursive evaluator so you could be seeing bad interactions there.

So should we expect to have duplicates when entering `simplifyFormula`?  And then freshen on the fly on places where it causes a crash?

> Which parser do you mean? The TIP parser freshens all variable names.

Yes my bad, I thought the expression above with duplicates came from the parser but in fact other transformations happened in between.

> Also, what do you mean about `freshenLocals` not being designed for certain variables?

I think it doesn't deduplicate, and transforms `(x$0 => x$0, x$0 => x$0)` into `(x$1 => x$1, x$1 => x$1)` rather than `(x$1 => x$1, x$2 => x$2)`? That's probably on purpose?

samarion:
> So should we expect to have duplicates when entering simplifyFormula? And then freshen on the fly on places where it causes a crash?

I don't think this is something we want to support. Ideally, we should find the places where duplication actually occurs and correctly freshen variables at that point.

The `RecursiveEvaluator` could also be fixed by either
1. implementing a recursive function for `Equals` in order to avoid the normalization step when evaluating lambda values, or
2. explicitly defining a value normalization function which is used only during `Equals` evaluation.

> I think it doesn't deduplicate, and transforms (x$0 => x$0, x$0 => x$0) into (x$1 => x$1, x$1 => x$1) rather than (x$1 => x$1, x$2 => x$2)? That's probably on purpose?

Oh, I didn't notice that. I don't think this is desired.

You can probably fix the behavior by using a [Transformer](https://github.com/epfl-lara/inox/blob/93503b1f7cffd68ae44483cdc2170cafba38eaa4/src/main/scala/inox/transformers/Transformer.scala#L6) with environment instead of storing a global mapping.

jad-hamza:
> I don't think this is something we want to support. Ideally, we should find the places where duplication actually occurs and correctly freshen variables at that point.

Ok thanks, I'll try to find where these come from.
 
> The `RecursiveEvaluator` could also be fixed by either
> 
>  1. implementing a recursive function for `Equals` in order to avoid the normalization step when evaluating lambda values, or
>  2. explicitly defining a value normalization function which is used only during `Equals` evaluation.

If I fix the duplication, is that still needed for that PR?

> You can probably fix the behavior by using a [Transformer](https://github.com/epfl-lara/inox/blob/93503b1f7cffd68ae44483cdc2170cafba38eaa4/src/main/scala/inox/transformers/Transformer.scala#L6) with environment instead of storing a global mapping.

I'm not sure what should be the behavior with `Choose` when `freshenChoose` is false:
https://github.com/epfl-lara/inox/blob/93503b1f7cffd68ae44483cdc2170cafba38eaa4/src/main/scala/inox/ast/ExprOps.scala#L85-L88

samarion:
> If I fix the duplication, is that still needed for that PR?

Well, that might be one of the sources of duplication (and you can't remove it currently because that would break `Equals` evaluation). I guess you could make sure to freshen locales before going to the solver, and that might be enough...

> I'm not sure what should be the behavior with Choose when freshenChoose is false.

The current behavior should  be preserved (don't freshen the choose variable). Chooses are a bit special, they behave more like a named function than a variable.

jad-hamza:
I fixed some duplications (pushed on this branch), and that fixed the `Lang.scala-0.tip` example but as you suggested this breaks other things. If I understand correctly, for now you suggest I keep the things that add duplication, but freshen with the new `freshenLocals` when entering `solveSAT` or other solver functions?

jad-hamza:
I removed the freshening that I had added during `RecursiveEvaluator` and fixed `freshenLocals`, and now things seem to work!

samarion:
> and now things seem to work!

Black magic :)

For `freshenLocals`, we ideally shouldn't do case matching on specific trees otherwise the method won't work on more complex Stainless trees. If you create the initial `env` based on `variablesOf(expr)`, you should be able to use the following pattern to generalize the transformation to arbitrary trees:

```scala
override def transform(e: Expr, env: Env): Expr = e match {
  case v: Variable => env(v)
  case _ =>
    val (..., vs, ...) = deconstructor.deconstruct(e)
    val newEnv = vs.foldLeft(env) { case (env, v) =>
      if (env contains v) env // probably should never happen but I'm not 100% sure
      else env + (v -> super.transform(v).freshen)
    }
    super.transform(e, newEnv)
}
```

I use `super.transform(v).freshen` when constructing `newEnv` in order to make sure we correctly modify potential variables in the dependent type of `v`.

Also, please preserve the `freshenChooses: Boolean = false` optional argument.

jad-hamza:
@samarion Should be good now, thanks for the help!
The PR for Stainless is ready as well: https://github.com/epfl-lara/stainless/pull/912/ (the build refers to Inox master so larabot on Stainless can only go through after this is merged)

src/main/scala/inox/ast/ExprOps.scala
```diff
@@ -65,30 +65,80 @@ trait ExprOps extends GenTreeOps {
...
   /** Freshens all local variables
-    * 
+    *
     * Note that we don't freshen choose ids as these are considered global
     * and used to lookup their images within models!
     */
-  def freshenLocals(expr: Expr, freshenChooses: Boolean = false): Expr = {
-    val subst: MutableMap[Variable, Variable] = MutableMap.empty
-    variablesOf(expr).foreach(v => subst(v) = v)
+  class Freshener(freshenChooses: Boolean = false) extends SelfTransformer {
```
samarion, line 72: Could you make this class and the two vals below protected to help indicate these are override points?

Also, you probably shouldn't use a default value for `freshenChooses` here to make sure sub-classes correctly propagate the constructor parameter.

src/main/scala/inox/ast/ExprOps.scala
```diff
@@ -65,30 +65,80 @@ trait ExprOps extends GenTreeOps {
...
+            val freshVd = vd.freshen
+            (sparams :+ transform(freshVd, env), env.updated(vd.id, freshVd.id))
+        }
+        SigmaType(sparams, transform(to, senv)).copiedFrom(tpe)
+
+      case _ => super.transform(tpe, env)
+    }
+  }
+
+  val freshenerNoChooses = new Freshener(false)
```
samarion, line 136: Micro-nit: I don't think you gain much by making these stable, I would suggest using a method instead to make the overrides in Stainless trees a bit nicer:
```scala
protected def freshener(freshenChooses: Boolean): Freshener = new Freshener(freshenChooses)
```

samarion (approved):

src/main/scala/inox/ast/ExprOps.scala
jad-hamza: Fixed, thanks! Should I even keep this function? (I called `new Freshener` directly from `freshenLocals`)

samarion: Since the implementation of `freshenLocals` is so simple, I guess there isn't really any point in having an intermediate function. Creating the `Freshener` directly and overriding `freshenLocals` in Stainless sounds good.

jad-hamza: Oh I thought overriding/extending the class would be enough but you're right I'll have to override `freshenLocals` (which is ok). It should be good now then, thanks!

PR #127, opened 2021-02-12
Use copy in toVariable to avoid sharing positions

jad-hamza:


src/main/scala/inox/ast/Definitions.scala
```diff
@@ -92,7 +92,7 @@ trait Definitions { self: Trees =>
     }
 
     /** Transform this [[ValDef]] into a [[Expressions.Variable Variable]] */
-    @inline def toVariable: Variable = v
+    @inline def toVariable: Variable = v.copy()
```
samarion, line 95: Should we add `copiedFrom(v)` to cover cases where the position is expected to stay the same?

jad-hamza: Yes good idea, thanks

PR #128, opened 2021-02-18
Allow Insecure Protocol to prevent warnings during build and upgrade to Scala 2.12.13 / sbt 1.3.0

jad-hamza:


build.sbt
```diff
@@ -31,9 +31,9 @@ unmanagedJars in Compile += {
 }
 
 resolvers ++= Seq(
-  "Sonatype OSS Snapshots" at "https://oss.sonatype.org/content/repositories/snapshots",
-  "Sonatype OSS Releases" at "https://oss.sonatype.org/content/repositories/releases",
-  "uuverifiers" at "http://logicrunch.research.it.uu.se/maven"
+  ("Sonatype OSS Snapshots" at "https://oss.sonatype.org/content/repositories/snapshots").withAllowInsecureProtocol(true),
+  ("Sonatype OSS Releases" at "https://oss.sonatype.org/content/repositories/releases").withAllowInsecureProtocol(true),
```
samarion, line 35: Is this change necessary? These repos seem to be using 'https' which isn't an insecure protocol.

samarion (commented):
You still need to update all the `.md` files under `src/main/doc` and replace "tut" with "mdoc".

Also, could you please check that adding a compilation error in the documentation correctly breaks compilation?

build.sbt
jad-hamza: No you're right, it's fixed now thanks.

jad-hamza:
> You still need to update all the `.md` files under `src/main/doc` and replace "tut" with "mdoc".
>
> Also, could you please check that adding a compilation error in the documentation correctly breaks compilation?

It all seems to compile properly now with `sbt mdoc` (I had to fix a few compilation errors).

I also added an extra case to silence a warning in normal Inox compilation:
https://github.com/epfl-lara/inox/pull/128/files#diff-e6b0bb85df4bd352bb54eb2b07207b020e5c04ed783e4df9982ccfcc83bf7d48R160

jad-hamza:
There are many warnings related to documentation links but I think there were there before?

doc/interpolations.md
```diff
@@ -44,26 +44,30 @@ import inox._
...
 ```scala
-scala> val tpe = t"Boolean"
-tpe: inox.trees.interpolator.trees.Type = Boolean
-
-scala> val expr = e"1 + 1 == 2"
-expr: inox.trees.interpolator.trees.Expr = 1 + 1 == 2
+val tpe = t"Boolean"
+// tpe: Type = Boolean
+val expr = e"1 + 1 == 2"
+// expr: Expr = Equals(
```
samarion, line 56: Do you know why pretty-printing is broken for expressions now?

samarion:
> There are many warnings related to documentation links but I think there were there before?

Yeah I don't think that's new. I guess we should go through the source code and fix all the broken links some day...

doc/interpolations.md
jad-hamza: Hmm no that's strange

jad-hamza:
> Yeah I don't think that's new. I guess we should go through the source code and fix all the broken links some day...

The goal is to have the links work when browsing on GitHub?

samarion:
> The goal is to have the links work when browsing on GitHub?

Aren't the warnings about broken links in the Scaladoc? If so, then we'd just get a bunch of broken links if we ever generated and published the Scaladoc.

If these are unrelated warnings, then I don't I know about them.

jad-hamza:
Actually these seem to be false positives (see https://github.com/scalameta/mdoc/issues/94), and the links work well on GitHub, see for instance:
https://github.com/jad-hamza/inox/blob/allow-insecure/doc/trees.md

samarion:
Hmm, I guess you can disable the link hygiene option: https://github.com/scalameta/mdoc/pull/129/files

jad-hamza:
Done, thanks!

doc/interpolations.md
samarion: It seems like we can use something like `mdoc:to-string` to use `toString()` instead of the weird custom printer: https://github.com/scalameta/mdoc/pull/148/files

jad-hamza: Thanks for looking this up! Should be good now

jad-hamza:
Awesome, thanks!

PR #129, opened 2021-02-22
No view

jad-hamza:
Same branch as https://github.com/epfl-lara/inox/pull/128 but without `.view` in `fold` to compare performance

### DIFF
--- a/src/main/scala/inox/ast/GenTreeOps.scala
+++ b/src/main/scala/inox/ast/GenTreeOps.scala
@@ -64,9 +64,7 @@ trait GenTreeOps { self =>
     val rec = fold(f) _
     val Deconstructor(es, _) = e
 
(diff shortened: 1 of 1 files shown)


jad-hamza:
The use of `.view` is making Stainless very slow on some examples, see:

http://laraquad4.epfl.ch:9000/epfl-lara/stainless/build/84
https://github.com/epfl-lara/stainless/commit/431c36e719e7951c18d80150f263923724bda062

One slow call for instance was from:
https://github.com/epfl-lara/stainless/blob/88093a7a88f497237c7b3cbdce4863f395ada88c/core/src/main/scala/stainless/extraction/innerfuns/Trees.scala#L183-L188

Removing `.view` fixes the issue:
http://laraquad4.epfl.ch:9000/epfl-lara/stainless/build/95
https://github.com/epfl-lara/stainless/pull/913

PR #131, opened 2021-03-31
Add option to remove colors and non-ascii characters, and to format errors for IDEs

jad-hamza:
Prepare support to address https://github.com/epfl-lara/stainless/issues/960


vkuncak:
@samarion could we merge this please?

vkuncak:
Green is good. Won't hold up matters with positions.

PR #133, opened 2021-04-14
Advise using jitpack.io resolver to depend on Inox

romac:
Given that there hasn't be a tagged release of Inox in ~2 years, use the latest commit (6dfb351eee) as the version.

One could alternatively specify `master-SNAPSHOT` to get the latest `master`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,17 +17,41 @@ information, see:
 - A more detailed description of the available solver/evaluator [API](doc/API.md) calls.
 - A description of the [trees](doc/trees.md) API and how to extend them.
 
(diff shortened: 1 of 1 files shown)


vkuncak (commented):
Can we additionally remind people: 

You can also depend on Inox by adding to your `build.sbt` code such as:
```scala
def ghProject(repo: String, version: String) = RootProject(uri(s"${repo}#${version}"))
lazy val inox = ghProject("https://github.com/epfl-lara/inox.git", "6dfb351eeee44a3f13152bf510aceba7936d0e4d")
```

Issue #115, opened 2020-01-08
Unsat file reported as SAT


jad-hamza:
```smt2
(define-fun f () (=> Int Int) (lambda ((loc Int)) 0))
(declare-datatypes () ((S (N) (C (r Int) (e Int)))))
(declare-datatypes () ((T (TT (body S)))))
(define-fun t () T (TT (C 22 5)))
(assert (not (let ((g f)) (= (@ (ite (is-C (body t)) (lambda ((r2 Int)) (ite (= r2 (r (body t))) (e (body t)) (@ g r2))) g) 22) 5))))
(check-sat)
```

```smt2
(define-fun f () (=> Int Int) (lambda ((loc Int)) 0))
(declare-datatypes () ((S (N) (C (r Int) (e Int)))))
(declare-datatypes () ((T (TT (body S)))))
(define-fun t () T (TT (C 22 5)))
(assert (not (= (@ (ite (is-C (body t)) (lambda ((r2 Int)) (ite (= r2 (r (body t))) (e (body t)) (@ f r2))) f) 22) 5)))
(check-sat)
```

The first file is reported as SAT (`--check-models` invalidates the model) while the second one (where the `let g` has been removed) is correctly reported as UNSAT.

The SMT file generated by Inox is:
```smt2
; Options: -in -smt2
(set-option :produce-unsat-assumptions true)
(declare-fun start!1 () Bool)
(assert start!1)
(declare-fun bs!1 () Bool)
(declare-fun b!8 () Bool)
(assert (= bs!1 (and b!8 start!1)))
(declare-fun lambda!5 () Int)
(declare-fun lambda!4 () Int)
(assert (=> bs!1 (not (= lambda!5 lambda!4))))
(assert (=> b!8 true))
(declare-fun order!1 () Int)
(declare-fun dynLambda!0 (Int Int) Int)
(assert (=> b!8 (< (dynLambda!0 order!1 lambda!4) (dynLambda!0 order!1 lambda!5))))
(declare-fun c!4 () Bool)
(assert (=> start!1 (= c!4 true)))
(declare-fun e!5 () Int)
(declare-fun dynLambda!1 (Int Int) Int)
(assert (=> start!1 (not (= (dynLambda!1 e!5 22) 5))))
(assert (=> b!8 (= e!5 lambda!5)))
(declare-fun b!9 () Bool)
(assert (=> b!9 (= e!5 lambda!4)))
(assert (= (and start!1 c!4) b!8))
[...]

PR #135, opened 2021-04-20
Implement @samarion's fix for issue 115 with lambda pointers

jad-hamza:
Fix issue #115

### DIFF
--- a/src/main/scala/inox/solvers/unrolling/Templates.scala
+++ b/src/main/scala/inox/solvers/unrolling/Templates.scala
@@ -514,7 +514,7 @@ trait Templates
 
   object Template {
 
-    def collectSelectors(expr: Expr, ptr: Expr): Seq[(Expr, Variable)] = expr match {
+    private def collectSelectors(expr: Expr, ptr: Expr): Seq[(Expr, Variable)] = expr match {
       case adt @ ADT(id, tps, es) => (adt.getConstructor.fields zip es).flatMap {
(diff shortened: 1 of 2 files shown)


src/main/scala/inox/solvers/unrolling/Templates.scala
```diff
@@ -514,23 +514,32 @@ trait Templates
 
   object Template {
 
-    def lambdaPointers(encoder: Expr => Encoded)(expr: Expr): Map[Encoded, Encoded] = {
-      def collectSelectors(expr: Expr, ptr: Expr): Seq[(Expr, Variable)] = expr match {
-        case adt @ ADT(id, tps, es) => (adt.getConstructor.fields zip es).flatMap {
-          case (vd, e) => collectSelectors(e, ADTSelector(ptr, vd.id))
-        }
+    def collectSelectors(expr: Expr, ptr: Expr): Seq[(Expr, Variable)] = expr match {
```
samarion, line 517: Nit: make it private?

samarion (commented):
Thanks for the fix! Could you also please add the broken example from the issue as a testcase?

src/main/scala/inox/solvers/unrolling/Templates.scala
jad-hamza: Sure! `resultPointers` as well?

jad-hamza:
> Thanks for the fix! Could you also please add the broken example from the issue as a testcase?

Thanks for writing the fix :) I've added the example

PR #136, opened 2021-04-20
Add support for smt-z3 with custom executable name

jad-hamza:
This should allow mixing several Z3 versions in portfolio mode: `--solvers=smt-z3,smt-z3:z3-4.8.10,smt-cvc4`

### DIFF
--- a/src/main/scala/inox/solvers/SolverFactory.scala
+++ b/src/main/scala/inox/solvers/SolverFactory.scala
@@ -65,14 +65,14 @@ object SolverFactory {
   import unrolling._
 
   val solverNames = Map(
-    "nativez3"     -> "Native Z3 with z3-templates for unrolling",
-    "nativez3-opt" -> "Native Z3 optimizer with z3-templates for unrolling",
-    "unrollz3"     -> "Native Z3 with inox-templates for unrolling",
-    "smt-cvc4"     -> "CVC4 through SMT-LIB",
-    "smt-z3"       -> "Z3 through SMT-LIB",
-    "smt-z3-opt"   -> "Z3 optimizer through SMT-LIB",
-    "smt-z3:EXEC"  -> "Z3 through SMT-LIB with custom executable name",
-    "princess"     -> "Princess with inox unrolling"
+    "nativez3"      -> "Native Z3 with z3-templates for unrolling",
+    "nativez3-opt"  -> "Native Z3 optimizer with z3-templates for unrolling",
+    "unrollz3"      -> "Native Z3 with inox-templates for unrolling",
+    "smt-cvc4"      -> "CVC4 through SMT-LIB",
+    "smt-z3"        -> "Z3 through SMT-LIB",
+    "smt-z3-opt"    -> "Z3 optimizer through SMT-LIB",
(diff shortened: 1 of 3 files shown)


src/main/scala/inox/solvers/SolverFactory.scala
```diff
@@ -213,25 +215,26 @@ object SolverFactory {
...
             val context = ctx
-          } with smtlib.Z3Solver {
+          } with smtlib.optimization.Z3Optimizer {
             val semantics: program.Semantics = targetSem
           }
         }
       })
 
-      case "smt-z3-opt" => create(p)(finalName, {
+      case _ if finalName.startsWith("smt-z3") => create(p)(finalName, {
```
samarion, line 224: I don't really like the fact that the case ordering becomes important. 

Could you maybe try with a more precise pattern?
```scala
case "smt-z3" | r"smt-z3:.+" => ...
```

jad-hamza: `r""` doesn't seem to work (does it require an import?), I ended up using:
```scala
case _ if finalName == "smt-z3" || finalName.startsWith("smt-z3:")
```

samarion: That's also fine.

src/main/scala/inox/solvers/SolverFactory.scala
```diff
@@ -109,6 +110,7 @@ object SolverFactory {
           replacement
 
         case Some(_) => name
+        case None if name.startsWith("smt-z3:") => name
```
samarion, line 113: Should we maybe check that the solver binary is available using
```scala
val hasZ3Exec = try {
  new Z3Interpreter(name.drop(7), Array("-in", "-smt2"))
  true
} catch {
  case _: java.io.IOException => false
}
```
and throw a fatalError if it's unavailable?

Also, maybe extract the `name.drop(7)` part into `getZ3Executable(name)`?

src/main/scala/inox/solvers/SolverFactory.scala
```diff
@@ -71,6 +71,7 @@ object SolverFactory {
     "smt-cvc4"     -> "CVC4 through SMT-LIB",
     "smt-z3"       -> "Z3 through SMT-LIB",
     "smt-z3-opt"   -> "Z3 optimizer through SMT-LIB",
+    "smt-z3:EXEC"  -> "Z3 through SMT-LIB with custom executable name",
```
samarion, line 74: Nit: maybe use `<exec>` instead of `EXEC` for consistency with the CVC4 options: https://github.com/epfl-lara/inox/blob/93550f1bcbf459e4c5584a5a87dbd0fbcb124eab/src/main/scala/inox/solvers/smtlib/CVC4Solver.scala#L13

src/main/scala/inox/solvers/SolverFactory.scala
jad-hamza: Sure, I've added this

src/main/scala/inox/solvers/SolverFactory.scala
jad-hamza: Fixed, thanks

PR #140, opened 2021-04-27
Add support for non-incremental mode

jad-hamza:
Thanks @samarion for the help

Built on top of https://github.com/epfl-lara/inox/pull/137

I wonder now if it wouldn't have been better to make new solver names such as `noinc:smt-z3` instead of one global option.
That way we could mix incremental and non-incremental solvers in one run.
But this will do for now, I'll that for a future PR.

### DIFF
--- a/src/it/scala/inox/tip/TipTestSuite.scala
+++ b/src/it/scala/inox/tip/TipTestSuite.scala
@@ -13,12 +13,14 @@ class TipTestSuite extends TestSuite with ResourceUtils {
     Seq(optSelectedSolvers(Set("nativez3")), optCheckModels(true)),
     Seq(optSelectedSolvers(Set("smt-z3")),   optCheckModels(true)),
     Seq(optSelectedSolvers(Set("smt-cvc4")), optCheckModels(true)),
-    Seq(optSelectedSolvers(Set("smt-z3")),   optCheckModels(true), optAssumeChecked(true))
+    Seq(optSelectedSolvers(Set("smt-z3")),   optCheckModels(true), optAssumeChecked(true)),
+    Seq(optSelectedSolvers(Set("smt-z3")),   optCheckModels(true), optNonIncremental(true))
   )
 
   override protected def optionsString(options: Options): String = {
     "solver=" + options.findOptionOrDefault(optSelectedSolvers).head +
-    (if (options.findOptionOrDefault(optAssumeChecked)) " assumechecked" else "")
+    (if (options.findOptionOrDefault(optAssumeChecked)) " assumechecked" else "") +
+    (if (options.findOptionOrDefault(optNonIncremental)) " --no-incremental" else "")
   }
 
   private def ignoreSAT(ctx: Context, file: java.io.File): FilterStatus = 
--- a/src/it/scala/inox/tip/TipTestSuite.scala
+++ b/src/it/scala/inox/tip/TipTestSuite.scala
@@ -10,17 +10,16 @@ import scala.language.existentials
 class TipTestSuite extends TestSuite with ResourceUtils {
 
   override def configurations = Seq(
-    Seq(optSelectedSolvers(Set("nativez3")), optCheckModels(true)),
-    Seq(optSelectedSolvers(Set("smt-z3")),   optCheckModels(true)),
-    Seq(optSelectedSolvers(Set("smt-cvc4")), optCheckModels(true)),
-    Seq(optSelectedSolvers(Set("smt-z3")),   optCheckModels(true), optAssumeChecked(true)),
-    Seq(optSelectedSolvers(Set("smt-z3")),   optCheckModels(true), optNonIncremental(true))
+    Seq(optSelectedSolvers(Set("nativez3")),      optCheckModels(true)),
+    Seq(optSelectedSolvers(Set("smt-z3")),        optCheckModels(true)),
+    Seq(optSelectedSolvers(Set("smt-cvc4")),      optCheckModels(true)),
+    Seq(optSelectedSolvers(Set("smt-z3")),        optCheckModels(true), optAssumeChecked(true)),
+    Seq(optSelectedSolvers(Set("no-inc:smt-z3")), optCheckModels(true))
   )
 
   override protected def optionsString(options: Options): String = {
     "solver=" + options.findOptionOrDefault(optSelectedSolvers).head +
-    (if (options.findOptionOrDefault(optAssumeChecked)) " assumechecked" else "") +
-    (if (options.findOptionOrDefault(optNonIncremental)) " --no-incremental" else "")
+    (if (options.findOptionOrDefault(optAssumeChecked)) " assumechecked" else "")
   }
 
   private def ignoreSAT(ctx: Context, file: java.io.File): FilterStatus = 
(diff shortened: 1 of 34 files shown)


src/it/scala/inox/tip/TipTestSuite.scala
```diff
@@ -13,12 +13,14 @@ class TipTestSuite extends TestSuite with ResourceUtils {
...
-    Seq(optSelectedSolvers(Set("smt-z3")),   optCheckModels(true), optAssumeChecked(true))
+    Seq(optSelectedSolvers(Set("smt-z3")),   optCheckModels(true), optAssumeChecked(true)),
+    Seq(optSelectedSolvers(Set("smt-z3")),   optCheckModels(true), optNonIncremental(true))
   )
 
   override protected def optionsString(options: Options): String = {
     "solver=" + options.findOptionOrDefault(optSelectedSolvers).head +
-    (if (options.findOptionOrDefault(optAssumeChecked)) " assumechecked" else "")
+    (if (options.findOptionOrDefault(optAssumeChecked)) " assumechecked" else "") +
+    (if (options.findOptionOrDefault(optNonIncremental)) " --no-incremental" else "")
```
samarion, line 23: For consistency, maybe drop the '--'

src/main/scala/inox/solvers/SolverFactory.scala
```diff
@@ -97,8 +97,40 @@ object SolverFactory {
...
 
+    val nonIncremental = ctx.options.findOptionOrDefault(optNonIncremental)
+    def nonIncrementalWrap[T, M](targetProgram: Program)(
+      nme: String,
+      targetSem: targetProgram.Semantics,
+      underlyingSolver: () => AbstractSolver {
+        val program: targetProgram.type
+        type Trees = T
+        type Model = M
+    }):
```
samarion, line 108: Micro nit: I would keep the type on the same line as the colon.

src/main/scala/inox/solvers/combinators/NonIncrementalSolver.scala
```diff
@@ -0,0 +1,49 @@
...
+  import program.trees._
+  import SolverResponses._
+
+  protected def underlying(): AbstractSolver {
+    val program: self.program.type
+    type Trees = self.Trees
+    type Model = self.Model
+  }
+
+  val assertions: IncrementalSet[Trees] = new IncrementalSet[Trees]
```
samarion, line 21: I know I mentioned `IncrementalSet` before, but maybe we should be using `IncrementalSeq` to keep the ordering deterministic (_e.g._ for easier debugging).

src/main/scala/inox/solvers/combinators/NonIncrementalSolver.scala
```diff
@@ -0,0 +1,49 @@
...
+  def reset(): Unit = assertions.clear()
+  def free(): Unit = for (solver <- allSolvers) solver.free()
+  def interrupt(): Unit = for (solver <- allSolvers) solver.interrupt()
+
+  override def check(config: CheckConfiguration): config.Response[Model, Assumptions] = {
+    val newSolver = underlying()
+    allSolvers.append(newSolver)
+    for (expression <- assertions)
+      newSolver.assertCnstr(expression)
+    newSolver.check(config)
```
samarion, line 35: Don't you think we should free the solver once it finishes `check`? Why are you keeping multiple solvers around?

src/main/scala/inox/Options.scala
```diff
@@ -232,3 +232,5 @@ object optSelectedSolvers extends SetOptionDef[String] {
 }
 
 object optNoColors extends inox.FlagOptionDef("no-colors", false)
+
+object optNonIncremental extends inox.FlagOptionDef("no-incremental", false)
```
samarion, line 236: Shouldn't this be somewhere under solvers?

jad-hamza: I ended up removing the option to use `no-inc:solver` instead

src/main/scala/inox/solvers/combinators/NonIncrementalSolver.scala
jad-hamza: You're right, can we assume that check's are called sequentially? So I keep an `Option[Solver]` to `free` or `interrupt` it if necessary, and then I remove it at the end of the check.

src/main/scala/inox/solvers/combinators/NonIncrementalSolver.scala
jad-hamza: Fixed, thanks

src/main/scala/inox/solvers/SolverFactory.scala
```diff
@@ -90,15 +90,27 @@ object SolverFactory {
   // extract <exec> in "smt-z3:<exec>"
   private def getZ3Executable(name: String): String = name.drop(7)
 
-  def getFromName(name: String, force: Boolean = false)
+  def getFromName(noIncName: String, force: Boolean = false)
```
samarion, line 93: I don't think it's very clean to change the API to include the no-inc detail. (I mean, I think keeping `name` is cleaner even though it means making more changes in the method body.)

src/main/scala/inox/solvers/SolverFactory.scala
```diff
@@ -324,7 +367,9 @@ object SolverFactory {
     }
   }
 
-  def supportedSolver(s: String) = solverNames.contains(s) || s.startsWith("smt-z3:")
+  private def supportedSolverRaw(s: String) = solverNames.contains(s) || s.startsWith("smt-z3:")
+  def supportedSolver(s: String) =
```
samarion, line 371: This isn't technically correct since only smt-z3, smt-cvc4 and unrollz3 allow no-inc

src/main/scala/inox/solvers/combinators/NonIncrementalSolver.scala
samarion: I'm not sure what you mean by sequentially. In general, I don't think we support multi-threading on the solvers. I would rather do something like
```scala
val newSolver = underlying()
try {
  ...
  newSolver.check(config)
} finally {
  newSolver.free()
}
```

jad-hamza: I also keep the solver in the `currentSolver` option to interrupt it, right?

samarion: Oooh, I hadn't realized you were keeping it around for interrupt :) my bad!

Yes, you should indeed keep it around in that case. Maybe just to keep things clean, you could throw an exception from `check` and `checkAssumptions` if the `currentSolver` isn't empty. (Or keep solvers in a list and add synchronization, but I don't see any reason to support multi-threading on this solver when it's not available on the others anyway.)

jad-hamza: I added an assertion, thanks

src/main/scala/inox/Main.scala
```diff
@@ -50,7 +51,8 @@ trait MainHelpers {
       "Use solvers s1,s2,...\nAvailable: " +
       solvers.SolverFactory.solverNames.toSeq.sortBy(_._1).map {
         case (name, desc) => f"\n  $name%-14s : $desc"
-      }.mkString("")
+      }.mkString("") +
+      "\nYou can prefix the solvers unrollz3, smt-z3, smt-z3:<exec>, smt-cvc4, with 'noinc:' to use them in non-incremental mode"
```
samarion, line 55: Isn't the prefix 'no-inc' (with a dash) ?

samarion: Also, maybe make it `and smt-cvc4` and remove the comma after `smt-cvc4`?

jad-hamza: Good catch, thanks

PR #141, opened 2021-04-28
Add model extraction for #unspecified and seq.unit

jad-hamza:


src/main/scala/inox/solvers/smtlib/SMTLIBParser.scala
```diff
@@ -106,6 +106,9 @@ trait SMTLIBParser {
     case SString(value) =>
       StringLiteral(utils.StringUtils.decode(value))
 
+    case FunctionApplication(QualifiedIdentifier(SimpleIdentifier(SSymbol("seq.unit")), None), Seq(SHexadecimal(hex))) =>
+      StringLiteral(utils.StringUtils.decode(hex.repr))
+
```
samarion, line 111: Isn't this a z3-specific representation? IIRC CVC4 uses a native representation for strings but Z3 uses `seq` internally. I'm not really sure anymore though.

jad-hamza: Yes I think you're right, we can always move it back here later if needed.

src/main/scala/inox/solvers/smtlib/SMTLIBParser.scala
```diff
@@ -224,7 +224,9 @@ trait SMTLIBParser {
       val MapType(from, to) = fromSMT(sort)
       FiniteMap(Seq.empty, d, from, to)
 
-    case _ => throw new MissformedSMTException(term, "Unknown SMT term")
+    case _ => throw new MissformedSMTException(term,
+      " Unknown SMT term of class: " + term.getClass
```
samarion, line 228: Nit: is the extra space before `Unknown` expected?

jad-hamza: No, I was concatenating something before it for debugging, thanks!

PR #142, opened 2021-04-29
Add let case in ChooseEncoder

jad-hamza:
Fix an issue that appeared on this Stainless program, where `ChooseEncoder` generates a `choose` function with an unbound variable.

```scala
iimport stainless.lang._
import stainless.lang.StaticChecks.WhileDecorations

object Bug {
  def insert[T](array: Array[T]): Unit = {
    val n: Int = array.length
    var i = n - 1

    (while (true) {
      i += 1

    }).inline.opaque.invariant (
      i <= n-1
    )
  }
}
```

### DIFF
--- a/src/main/scala/inox/solvers/unrolling/ChooseEncoder.scala
+++ b/src/main/scala/inox/solvers/unrolling/ChooseEncoder.scala
@@ -25,13 +25,9 @@ trait ChooseEncoder extends transformers.ProgramTransformer {
 
     val newFds = sourceProgram.symbols.functions.values.toList.map { fd =>
       def rec(e: Expr, params: Seq[ValDef]): Expr = e match {
-        case l: Let =>
-          val free = exprOps.variablesOf(l)
-          l.copy(body = rec(l.body, params.filter(vd => free(vd.toVariable)) :+ l.vd)).copiedFrom(l)
+        case l: Let => l.copy(body = rec(l.body, params :+ l.vd)).copiedFrom(l)
 
-        case l: Lambda =>
-          val free = exprOps.variablesOf(l)
-          l.copy(body = rec(l.body, params.filter(vd => free(vd.toVariable)) ++ l.params)).copiedFrom(l)
+        case l: Lambda => l.copy(body = rec(l.body, params ++ l.params)).copiedFrom(l)
 
         case c: Choose =>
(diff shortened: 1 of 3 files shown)


src/main/scala/inox/solvers/unrolling/UnrollingSolver.scala
```diff
@@ -164,8 +164,15 @@ trait AbstractUnrollingSolver extends Solver { self =>
...
   def assertCnstr(expression: Expr): Unit = context.timers.solvers.assert.run(try {
     context.timers.solvers.assert.sanity.run {
-      symbols.ensureWellFormed // make sure that the current program is well-formed
-      typeCheck(expression, BooleanType()) // make sure we've asserted a boolean-typed expression
+      try {
+        symbols.ensureWellFormed // make sure that the current program is well-formed
+        typeCheck(expression, BooleanType()) // make sure we've asserted a boolean-typed expression
+      } catch {
+        case _: Throwable =>
+          context.reporter.internalError(
```
samarion, line 172: Can't we include the original error in the exception? This will make debugging harder since we'll lose the original stack trace.

src/main/scala/inox/solvers/unrolling/ChooseEncoder.scala
```diff
@@ -25,6 +25,10 @@ trait ChooseEncoder extends transformers.ProgramTransformer {
 
     val newFds = sourceProgram.symbols.functions.values.toList.map { fd =>
       def rec(e: Expr, params: Seq[ValDef]): Expr = e match {
+        case l: Let =>
+          val free = exprOps.variablesOf(l)
+          l.copy(body = rec(l.body, params.filter(vd => free(vd.toVariable)) :+ l.vd)).copiedFrom(l)
```
samarion, line 30: After a bit more thought, I think we should probably remove the filtering logic here (and in the `Lambda` case). I don't think it makes sense to restrict the variables chooses under lets/lambdas have access to.

src/main/scala/inox/solvers/unrolling/UnrollingSolver.scala
samarion: After some more thought, I think it would make more sense to do the try-catch inside Stainless. This gives more flexibility to Inox clients about how they want to handle/display the error.

PR #143, opened 2021-04-30
Add infoContinue method in Reporter to emit without line return

jad-hamza:
This is to get the following output in Stainless, where each `.` corresponds to a valid VC.

```scala
[  Info  ] Starting verification..........................
[Warning ] SortedArray.scala:507:14:  => TIMEOUT
                 assert(isSortedRange(array, order, i+2, array.length-1))
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
^C[  Info  ] 
[Warning ] Interrupted...
[Warning ] SortedArray.scala:508:14:  => CANCELLED
                 assert(order.leq(array(i+1), array(i+2)))
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[  Info  ] ......................................................
[Warning ] SortedArray.scala:508:42:  => TIMEOUT
                 assert(order.leq(array(i+1), array(i+2)))
                                                    ^^^
[  Info  ] ..........................................................................................................................................................................................................................
```

### DIFF
--- a/src/main/scala/inox/Reporter.scala
+++ b/src/main/scala/inox/Reporter.scala
@@ -40,6 +40,8 @@ abstract class Reporter(val debugSections: Set[DebugSection]) {
 
   def emit(msg: Message): Unit
 
+  def infoContinue(msg: Any): Unit
+
   def onFatal(msg: String = ""): Nothing = throw FatalError(msg)
 
   def onCompilerProgress(current: Int, total: Int) = {}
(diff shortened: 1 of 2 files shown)


samarion:
I wonder if we couldn't use the `onCompilerProgress` method instead? You can rename it to `onProgress` if you prefer, and then you can override it in Stainless to implement this behavior.

jad-hamza:
Ah good idea I didn't know about that. Apparently the function is already used elsewhere, this creates output before checking the verification conditions.

jad-hamza:
I removed this https://github.com/epfl-lara/stainless/blob/a90bc66cd1a7601f5776f555be290a25f45f8253/frontends/scalac/src/main/scala/stainless/frontends/scalac/ScalaCompiler.scala#L127 and now it's ok; I'll close here and open a PR in Stainless.

PR #149, opened 2021-05-09
Avoid using checkAssumptions in non-incremental mode

jad-hamza:
See: https://github.com/Z3Prover/z3/issues/5257


src/main/scala/inox/solvers/combinators/NonIncrementalSolver.scala
```diff
@@ -52,7 +52,9 @@ trait NonIncrementalSolver extends AbstractSolver { self =>
       currentSolver = Some(newSolver)
       for (expression <- assertions)
         newSolver.assertCnstr(expression)
-      val res = newSolver.checkAssumptions(config)(assumptions)
+      for (assumption <- assumptions)
```
samarion, line 55: Please add a comment about why this is necessary (pointing to the Z3 bug).

PR #151, opened 2021-05-11
Make simplestValue top-level for overrides + hasInstance fix on ADTType

jad-hamza:


samarion:
Not super important, but might make sense to split the tuple fix into a second PR. Did you check that it fixes the issue on the Stainless side?

jad-hamza:
> Did you check that it fixes the issue on the Stainless side?

Yes and I encountered https://github.com/epfl-lara/inox/issues/153 on the same example, I think this means it was OK for the tuples.

PR #155, opened 2021-05-18
Merge rust interop features

### LINKED ISSUE #152: Integration of `rust-interop` in `master`
Landing the `rust-interop` features on `master` would have enormous benefits for the [epfl-lara/rust-stainless](https://github.com/epfl-lara/rust-stainless) project in terms of less maintenance for getting updates. This is under the condition that the features added/changed on the rust-interop branch can co-exist with the current master.

@samarion suggested the following changes to `rust-interop` before a potential merge:

- [x] I would add an option optFormat which can be either “tip” or “binary”
- [x] move deserializerMain into the Main object unless you need it in Stainless (and same for the optDeserialize option)
- [x] Then I’d create a parsing method in the Main with return type (Program, Option[Expr]) (or maybe Symbols instead of Program, depending on the API of the TIP parser’s parseScript method)
- [x] If the format option is “tip”, then call the current TIP parsing logic and return the resulting program and Some of the expression
- [x] If the format option is “binary”, deserialize the input and check if there’s a boolean method with some pre-defined name in the program, e.g. check. Then return the deserialized program (minus the check method) and the check method’s body as the expression. If no such method exists in the program, return None for the Option[Expr].
- [x] Then, if the returned option is Some, run solving on that expression.
- [x] In addition, I would add a boolean option optPrintProgram which, if true, has Inox print the program (and potential expression) returned by either the “tip” or “binary” parser.

yannbolliger:
Closes #152.

This PR requests to merge _all_ Rust interoperability features into `master` of Inox.
The changes requested by @samarion were implemented, there is now an input format called `binary`.


src/main/scala/inox/utils/RustInteropGenerator.scala
```diff
@@ -0,0 +1,404 @@
...
+
+package inox
+package utils
+
+import java.io.{File, FileWriter}
+
+import scala.reflect.runtime.universe._
+import scala.collection.mutable.{ArrayBuffer, HashMap => MutableMap, HashSet => MutableSet}
+
+trait RustInteropGeneration { self: InoxSerializer =>
```
samarion, line 11: I hadn't realized this was going to live inside Inox. Is there any point having it here rather than in Stainless? I'd prefer if Inox stayed mostly frontend-independent (if possible).

samarion:
You might want to add yourself to [.larabot.conf](https://github.com/epfl-lara/inox/blob/master/.larabot.conf) (in a separate PR) so that the CI can run on your PRs.

yannbolliger:
> You might want to add yourself to [.larabot.conf](https://github.com/epfl-lara/inox/blob/master/.larabot.conf) (in a separate PR) so that the CI can run on your PRs.

Yes, that would be great: https://github.com/epfl-lara/inox/pull/156

src/main/scala/inox/utils/RustInteropGenerator.scala
yannbolliger: I understand that. We absolutely need this file to generate our AST trees in Rust but as long as it still works, nothing speaks against moving this to Stainless. Do you think everything will still be accessible there?

yannbolliger: OK, I tested it and there is no problem if we move this entire file to Stainless. See here: https://github.com/epfl-lara/stainless/pull/1060

samarion (approved):
I think you'll need to push a new commit for larabot to pick up the PR. Maybe rebase on origin/master and force-push?

src/main/scala/inox/Main.scala
```diff
@@ -215,20 +246,25 @@ trait MainHelpers {
...
+    for {
+      file <- getFilesOrExit(ctx)
+      (program, exprOpt) <- parse(ctx, file)
+    } {
+      import ctx._
+      import program._
 
+      if (ctx.options.findOptionOrDefault(optPrintProgram)) {
+        ctx.reporter.info(s"Program in $file:\n\n")
+        ctx.reporter.info(program)
```
samarion, line 264: I forgot to mention in my previous review, you might want to use `program.asString(PrinterOptions.fromContext(ctx))` here so that printer options are correctly taken into account.

yannbolliger: Done 👍

PR #158, opened 2021-06-22
When shifting right, only add ones for signed bitvectors

jad-hamza:


jad-hamza:
@samarion can you have a look please? (this is for `--eval` in Stainless)

jad-hamza:
Thanks :)

PR #161, opened 2021-08-26
Move interrupted check lower in NonIncrementalSolver to avoid race conditions

jad-hamza:
@samarion I'm moving the interrupt check a bit later to avoid the case where we could get interrupted right after the `if` check.

### DIFF
--- a/src/main/scala/inox/solvers/combinators/NonIncrementalSolver.scala
+++ b/src/main/scala/inox/solvers/combinators/NonIncrementalSolver.scala
@@ -29,38 +29,48 @@ trait NonIncrementalSolver extends AbstractSolver { self =>
   var interrupted = false
 
   def interrupt(): Unit = {
-    interrupted = true
-    for (solver <- currentSolver) solver.interrupt()
+    val optSolver = synchronized {
+      interrupted = true
+      currentSolver
+    }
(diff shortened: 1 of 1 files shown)


jad-hamza:
Which one do you think? I was having a look at `AtomicReference`, and store both the `currentSolver` and `interrupt` in a pair here, but I don't think it's ideal when you want to set just one component.

samarion:
I think we can just use a `synchronized` block and make sure modifications of `interrupted` and `currentSolver` are atomic. Something like:
```scala
def interrupt(): Unit = {
  val optSolver = synchronized {
    interrupted = true
    currentSolver
  }
  for (solver <- optSolver) solver.interrupt()
}

override def check(config: CheckConfiguration): config.Response[Model, Assumptions] = {
  assert(currentSolver.isEmpty, ...)
  val newSolver = underlying()
  try {
    val runCheck = synchronized {
      currentSolver = Some(newSolver)
      !interrupted
    }
    if (runCheck) {
      ...
```

jad-hamza:
> I think we can just use a `synchronized` block and make sure modifications of `interrupted` and `currentSolver` are atomic. 

Thanks, should be good now

jad-hamza:
Thanks :)

PR #163, opened 2021-09-03
Add debug option to dump blocker graph

gsps:
Adds a debug section `blocker-graph` that dumps the (non-strict) implication graph among all blocking literals. The current graph is dumped at the end of `checkAssumptions` and saved in `blocker-graphs/*file*-*n*.dot`.

Nodes that correspond to `defBlocker`s also contain the corresponding function name and the call site's position. The latter is information is unreliable, since we only store the call site position when the callee's `FunctionTemplate` is first cached (the cache key being the `TypedFunDef`).

@samarion Any preference on where to move the dot-graph-dumping code that new lives in `checkAssumptions`?

### DIFF
--- a/src/main/scala/inox/solvers/unrolling/FunctionTemplates.scala
+++ b/src/main/scala/inox/solvers/unrolling/FunctionTemplates.scala
@@ -33,12 +33,13 @@ trait FunctionTemplates { self: Templates =>
   object FunctionTemplate {
     private val cache: MutableMap[TypedFunDef, FunctionTemplate] = MutableMap.empty
 
-    def apply(tfd: TypedFunDef): FunctionTemplate = cache.getOrElse(tfd, {
+    def apply(tfd: TypedFunDef, pos: utils.Position): FunctionTemplate = cache.getOrElse(tfd, {
       val body: Expr = timers.solvers.simplify.run { simplifyFormula(tfd.fullBody) }
 
       val tpVars = tfd.tps.flatMap(variableSeq).distinct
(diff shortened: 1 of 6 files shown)


src/main/scala/inox/solvers/unrolling/FunctionTemplates.scala
```diff
@@ -33,12 +33,13 @@ trait FunctionTemplates { self: Templates =>
...
 
-    def apply(tfd: TypedFunDef): FunctionTemplate = cache.getOrElse(tfd, {
+    def apply(tfd: TypedFunDef, pos: utils.Position): FunctionTemplate = cache.getOrElse(tfd, {
       val body: Expr = timers.solvers.simplify.run { simplifyFormula(tfd.fullBody) }
 
       val tpVars = tfd.tps.flatMap(variableSeq).distinct
       val fdArgs: Seq[Variable] = tfd.params.map(_.toVariable)
-      val call: Expr = tfd.applied(fdArgs)
+      // Note: This position is best-effort, since function templates are cached.
+      val call: Expr = tfd.applied(fdArgs).setPos(pos)
```
samarion, line 42: Is the goal here to ensure that all calls in the dumped blocker graph have positions? If so, I don't think you need to set a position on this call because it should never be part of the blocker graph.

(If possible, I'd like to avoid making positions part of the `FunctionTemplate` API)

samarion (commented):
Maybe you could add an `UnrollingDebugger` trait similarly to the `SMTLIBDebugger` and have it override `checkAssumptions`?

PR #170, opened 2021-11-18
Adding BitVectors literals arithmetic simplification

mario-bucev:
Note: targets the `scala-3.x` branch.
That addition is needed for some tests in `GenC` (with the Dotty frontend) to succeed (more specifically, some arrays require a literal size). The Stainless extraction happens before the constant unfolding in the Dotty pipeline.
The other two minor changes are here to help with negative tests checks (for later on)

### DIFF
--- a/src/main/scala/inox/ast/ExprOps.scala
+++ b/src/main/scala/inox/ast/ExprOps.scala
@@ -226,41 +226,80 @@ class ExprOps private(val trees: Trees)
     * that reduce the size of the tree. The only guarantee from this function is
     * to not augment the size of the expression and to be sound.
     */
-  def simplifyArithmetic(expr: Expr): Expr = {
+  def simplifyArithmetic(expr: Expr)(using Symbols): Expr = {
+    // Extractor for BigInt and BV literals.
+    sealed trait ExIntegerLit {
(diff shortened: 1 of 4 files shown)


src/main/scala/inox/ast/ExprOps.scala
```diff
@@ -227,40 +227,81 @@ class ExprOps private(val trees: Trees)
...
+      case Division(e, exLit(_, one, _)) if one == BigInt(1) => Some(e)
+
+      case _ => None
+    }
+    val bigIntSimplifier = integerArithSimp(ExBigIntLit, (_, l, _) => IntegerLiteral(l))
+    val bvSimplifier = integerArithSimp(ExBVLit, BVLiteral.apply)
+
+    object ExIntegerArithSimp {
+      def unapply(e: Expr): Option[Expr] =
+        bigIntSimplifier(e).orElse(bvSimplifier(e))
```
samarion, line 300: Maybe you could first check whether `e` has an integer or bitvector type to dispatch to the right simplifier. That would also allow you to drop all the sign and size checks since you would know the expression is well-typed.

mario-bucev: Good point!

PR #175, opened 2022-01-07
Upgrade ScalaZ3 to 4.8.14

mario-bucev:


src/main/scala/inox/solvers/z3/Z3Native.scala
```diff
@@ -654,7 +654,26 @@ trait Z3Native extends ADTManagers with Interruptible { self: AbstractSolver =>
...
+              case StringType() if z3.getDeclKind(decl) == OpSeqEmpty =>
+                assert(args.isEmpty)
+                StringLiteral("")
+
+              case StringType() if z3.getDeclKind(decl) == OpSeqUnit =>
+                assert(args.size == 1)
+                rec(args.head, BVType(false, 16), seen) match {
+                  case lit@BVLiteral(_, _, _) =>
+                    StringLiteral(String.valueOf(lit.toBigInt.toChar))
+                  case rarg => rarg
```
samarion, line 666: What case are we expecting this to match? Won't we run into type issues by returning something with expected type `BVType(false, 16)` as a string?

src/main/scala/inox/solvers/z3/Z3Native.scala
```diff
@@ -852,11 +871,11 @@ trait Z3Native extends ADTManagers with Interruptible { self: AbstractSolver =>
     val chooses: MutableMap[(Identifier, Seq[Type]), Expr] = MutableMap.empty
     chooses ++= ex.chooses.map(p => (p._1, Seq.empty[Type]) -> p._2)
 
-    chooses ++= model.getFuncInterpretations.flatMap { case (decl, mapping, _) =>
+    chooses ++= model.getFuncInterpretations.flatMap { case (decl, mapping, defaultValue) =>
       functions.getA(decl).flatMap(tfd => tfd.fullBody match {
         case c: Choose =>
           val ex = new ModelExtractor(model)
-          val body = mapping.foldRight(c: Expr) { case ((args, res), elze) =>
+          val body = mapping.foldRight(ex(defaultValue, tfd.getType)) { case ((args, res), elze) =>
```
samarion, line 878: Good catch!

src/main/scala/inox/solvers/z3/Z3Unrolling.scala
```diff
@@ -68,7 +68,11 @@ trait Z3Unrolling extends AbstractUnrollingSolver { self =>
   private case class ModelWrapperImpl(model: Z3Model) extends super.ModelWrapper {
     private val ex = new underlying.ModelExtractor(model)
 
-    def extractConstructor(v: Z3AST, tpe: t.ADTType): Option[Identifier] = model.eval(v).flatMap {
+    private def tryZ3[T](res: => T): Option[T] =
+      // @nv: Z3 sometimes throws an exception when functions are called after Z3 has been canceled
+      try { Some(res) } catch { case e: Z3Exception if e.getMessage == "canceled" => None }
```
samarion, line 73: Nit: should this rather be something like `tryEval`? It seems to be always used as a wrapper for `model.eval(...)`.

src/main/scala/inox/solvers/SolverResponses.scala
```diff
@@ -99,11 +99,14 @@ object SolverResponses {
     }).asInstanceOf[Response[M,C]]
 
     def convert[M1,T1,M2,T2](resp: Response[M1,Set[T1]], cm: M1 => M2, cc: Set[T1] => Set[T2]): Response[M2,Set[T2]] =
+      convertMaybe(resp, (m1: M1) => Some(cm(m1)), cc)
+
+    def convertMaybe[M1,T1,M2,T2](resp: Response[M1,Set[T1]], cm: M1 => Option[M2], cc: Set[T1] => Set[T2]): Response[M2,Set[T2]] =
```
samarion, line 104: It seems to me like there's some confusion with where the `tryZ3` logic is invoked. Some calls to `model.eval` will be wrapped in a try-catch while others won't be so it's unclear whether we'll end up with `Unknown` or a partial model after the call to `convertMaybe`.

I think it would make more sense to keep only the `convert` call and make sure to wrap all calls into the `Z3Model` class with `tryZ3` in `Z3Native.extractModel`. We can then adapt the `UnrollingSolver` to always check if search was interrupted _after_ having converted the model and return `Unknown` in that case (since the model will be partial): https://github.com/epfl-lara/inox/blob/2f402f9bf9541cc061558e525fdd53efa5f86a87/src/main/scala/inox/solvers/unrolling/UnrollingSolver.scala#L589-L590

samarion (commented):
These all look like good changes :)

src/main/scala/inox/solvers/z3/Z3Native.scala
mario-bucev: You are right, I did not think about that. I guess matching that `rarg` branch will make a fine test case :p

src/main/scala/inox/solvers/z3/Z3Unrolling.scala
mario-bucev: Good suggestion!

src/main/scala/inox/solvers/SolverResponses.scala
mario-bucev: I like that approach; scattering `tryZ3` a bit everywhere is indeed not very clean :p

src/main/scala/inox/solvers/z3/Z3Native.scala
mario-bucev: I have looked at how `Z3Target` handles that similar case, and it appears that it expects a (char) literal. It seems we cannot do much, so I've put a call to `unsound`.

src/main/scala/inox/solvers/z3/Z3Unrolling.scala
samarion: I see we now have a bunch of different implementations of `tryZ3` variants. Could we consolidate them all into a single implementation in `Z3Native.scala`? You can create a companion object if it makes more sense.

You can also provide a `tryZ3Opt` if you prefer that rather than calling `.flatten` at `tryZ3` call-sites.

src/main/scala/inox/solvers/SolverResponses.scala
samarion: I think you're mapping to `Unknown` too early. We need to make sure that the model extraction performed by `CheckResult.cast` never results in a partial model. In your implementation, if an interruption occurred between the solver result and extraction, we could still end up with an incomplete model.

I believe the logic should look something like:
```scala
def cast(...) = {
  val res = config.convert(config.cast(resp), extractSimpleModel, decodeAssumptions)
  if (interrupted) Unknown else res
}
```

mario-bucev: Good catch, thanks!

mario-bucev: Hopefully the pushed change correctly addresses that point!

samarion: Looking good!

Sorry I realized I missed a similar check in the `Validate(model)` case: https://github.com/epfl-lara/inox/blob/822692640f458a81f48de79975269fbca261dc61/src/main/scala/inox/solvers/unrolling/UnrollingSolver.scala#L690

Maybe it suffices to move the check into the `CheckResult.apply` method (and only if `checkModels` is false the the model will actually be evaluated otherwise), something like:
```scala
object CheckResult {
  def cast(resp: SolverResponse[underlying.Model, Set[underlying.Trees]]): CheckResult =
    CheckResult(config.convert(config.cast(resp), extractSimpleModel, decodeAssumptions))

  def apply[M <: Model, A <: Assumptions](resp: config.Response[M, A]) = resp match {
    case SatWithModel(model) if !checkModels && abort => new CheckResult(config cast Unknown)
    case _ => new CheckResult(resp)
  }

  def unapply(res: CheckResult): Option[config.Response[Model, Assumptions]] = Some(res.response)
}
```

Note that `abort` is basically the same as `context.interruptManager.isInterrupted`.

samarion: Sometimes the solvers return a model despite/when they get interrupted, so we can still use these when they appear if we run `checkModels` since the check ensures the model is indeed valid.

mario-bucev: Good to know, thanks for the explanation :)

mario-bucev:
Thanks for taking the time! I sadly did not squash the commits with the last commit, ah well...

PR #181, opened 2022-05-17
Make PurityOptions a case class, fix warnings, disable scaladoc gen

mario-bucev:
Note: `PurityOptions` is used as a key in some maps, having it as a non-case class leads to unexpected results (due to the nature of `hash` and `==` relying on reference equality instead of structural equality)


mario-bucev:
We can also use this PR to update the Scala-SMTLIB dependency once https://github.com/epfl-lara/scala-smtlib/pull/8 is merged

PR #184, opened 2022-09-26
Adding a case for type instantiation

mario-bucev:
Concerns the case `ADTType -> RefinementType`. However, we drop the `prop` part, so maybe this is incorrect?

### DIFF
--- a/src/main/scala/inox/ast/TypeOps.scala
+++ b/src/main/scala/inox/ast/TypeOps.scala
@@ -72,6 +72,15 @@ trait TypeOps {
         (ts1 zip ts2).foldLeft[Instantiation](Map.empty) {
           case (inst, (tp1, tp2)) => unify(inst, rec(tp1, tp2))
         }
+      case (adt1: ADTType, refn: RefinementType) =>
+        refn.getType match {
+          case adt2: ADTType =>
+            // ADT(X1, ..., Xn) -> { vd: ADT(A1, ..., An) | prop }
+            // ~>
+            // X1 -> A1, ..., Xn -> An
+            rec(adt1, adt2)
+          case _ => throw new Failure
+        }
       case _ => throw new Failure
     }
 


src/main/scala/inox/ast/TypeOps.scala
```diff
@@ -72,6 +72,15 @@ trait TypeOps {
...
         }
+      case (adt1: ADTType, refn: RefinementType) =>
+        refn.getType match {
+          case adt2: ADTType =>
+            // ADT(X1, ..., Xn) -> { vd: ADT(A1, ..., An) | prop }
+            // ~>
+            // X1 -> A1, ..., Xn -> An
+            rec(adt1, adt2)
+          case _ => throw new Failure
+        }
```
samarion, line 83: I think you could make it a bit more general with:
```scala
case (_, RefinementType(vd, _)) =>
  rec(from, vd.tpe)
```

Also, we probably need cases for `SigmaType` and `PiType`:
```scala
case (TupleType(tps), SigmaType(vds, elem)) =>
  val vars = vds.map(_.toVariable).toSet
  (tps zip (vds.map(_.tpe) +: elem)).foldLeft[Instantiation](Map.empty) {
    case (inst, (tp1, tp2)) =>
      unify(inst, rec(tp1, if ((exprOps.variablesOf(tp2) & vars).nonEmpty) tp2.getType else tp2))
  }
```
(and same for `PiType`).


Finally, you should probably make sure the `from` type isn't a dependent type by using `from.getType` in the call to `rec` below.

vkuncak:
Is there a test case for this, @mario-bucev ?

src/main/scala/inox/ast/TypeOps.scala
mario-bucev: Thank you for the suggestions! I'll go back to this PR afterwards :)

vkuncak:
So it's too cumbersome to generate tests at the level of Inox?

mario-bucev:
We should be able to unit test this part

PR #201, opened 2023-10-04
Add sbt assembly to create self-contained jars

vkuncak:


mario-bucev (approved):
Works like a charm :)

PR #203, opened 2023-10-10
Add support for cvc5

mario-bucev:
cvc5 can be selected with `--solvers=smt-cvc5`, provided that its executable (named `cvc5`) is found in `PATH`.


vkuncak:
Looks like changes from cvc4 are small! Look forward to making use of more cvc5 functionality in the future.

mario-bucev:
Indeed, it was a pleasant surprise :)

PR #204, opened 2023-10-31
Improve performance of DiGraph

mario-bucev:
Copied/pasted comment from `Graphs.scala`:
Implementation note: `inEdgesMap` and `outEdgesMap` can be derived from `N` and `E`.
However, since `inEdges` and `outEdges` are often used with updated copies of the graph, it is interesting performance-wise to have these maps as fields and "update" them alongside "update" operation such as `+`, `++` etc.

### DIFF
--- a/src/it/scala/inox/TestSuite.scala
+++ b/src/it/scala/inox/TestSuite.scala
@@ -8,7 +8,7 @@ import org.scalatest.matchers.should.Matchers
 import org.scalatest.{CancelAfterFailure, Tag, exceptions}
 import utils._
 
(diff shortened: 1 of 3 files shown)


samarion:
Are you sure this improves performance in the Inox/Stainless end-to-end flows?

Usually we construct the entire graph before we start looking into transitive edges, so computing the fields for intermediate graphs might actually be slower.

mario-bucev:
It does for large programs. For instance loading `verification/valid` is significantly faster in the preprocessing step. Of course this may have an impact for smaller programs.

PR #206, opened 2024-04-12
Add a function to get the SMT lib ID from the solver if needed

samuelchassot:


samuelchassot:
For example, it is used to read the SMT-Lib ID from the verification checker of Stainless, to be able to display it in the summary table.

vkuncak (approved):
Looks reasonable and should only affect smt-lib debugging.

Issue #210, opened 2024-05-07
Inox does not compile on Scala 3.3.1 +


sankalpgambhir:
Inox is currently on Scala 3.3.0, and fails to compile on 3.3.1 or higher. I noticed this while updated my build to update dependencies generally.

The following error is raised:

```scala
[error] -- Error: <working_dir>/src/main/scala/inox/solvers/SolverFactory.scala:207:14 
[error] 207 |        class NativeZ3OptImpl(override val program: p.type)
[error]     |              ^
[error]     |parent trait NativeZ3Optimizer has a super call which binds to the value inox.solvers.unrolling.AbstractUnrollingSolver.targetProgram. Super calls can only target methods.
[error] one error found
```

The error seems to be due to a pattern that was disallowed in Scala 3.3.1 ([scala/scala3#16908](https://www.github.com/scala/scala3/pull/16908)) as in the following minimal example:

```scala
trait A              { def f: String }
trait B extends A    { def f = "B" }
trait C extends A    { override val f = "C" }
trait D extends A, B { def d = f }
trait E extends A, B { def d = super.f }

class O1 extends A, B, C, D // passes (why though?)
class O2 extends A, B, C, E // fails
```

A `val` overrides a `def` in a parent trait, and there is an ambiguous access through a subclass. I don't get why leaving out the `super.` causes it to work.

In any case, despite the error message, a simple search shows that the Inox code base has no instance of `def targetProgram` or `super.targetProgram`, so it's not obvious where this is coming from. I do not see any other clear `super` calls accessing `val`s either.  It would need some more investigation. 

This also seems to only happen for that single instance (and commenting it out leads to a successful compile) suggesting that the error is isolated, and not a more global pattern problem. 

If anyone else has ideas about what is causing this, it would be helpful to hear them.

samuelchassot:
https://github.com/epfl-lara/inox/pull/209

this is discussed here in my PR.

samuelchassot:
> This also seems to only happen for that single instance (and commenting it out leads to a successful compile) suggesting that the error is isolated, and not a more global pattern problem.
This is good news!

sankalpgambhir:
Oh thanks, I was not in the loop with the Stainless PR! I mostly wanted to leave this here so I have a documented source and don't forget about the details. I will keep passively looking into it while I'm working on Inox.

samuelchassot:
NP! Okay, sounds good! Thanks :)

sankalpgambhir:
[...]

PR #211, opened 2024-05-13
Bump to Scala 3.3.3 with small fix

sankalpgambhir:
Bump Inox to Scala 3.3.3 LTS and sbt to 1.10.0.

On bumping the version, there is a compilation error during type checking (see #210 for details). In short, it boils down to the following declaration:

https://github.com/epfl-lara/inox/blob/9fdb4d494133595fd832584e8aaa64b3f027fbab/src/main/scala/inox/solvers/z3/NativeZ3Optimizer.scala#L21-L23

where the reference to `targetProgram.type` raises the error:

```scala
[error] -- Error: <working_dir>/src/main/scala/inox/solvers/SolverFactory.scala:207:14 
[error] 207 |        class NativeZ3OptImpl(override val program: p.type)
[error]     |              ^
[error]     |parent trait NativeZ3Optimizer has a super call which binds to the value inox.solvers.unrolling.AbstractUnrollingSolver.targetProgram. Super calls can only target methods.
[error] one error found
```

Weirdly, the error is simply fixed by adding an alias:

https://github.com/sankalpgambhir/inox/blob/9144039a31d401653af09671e4a58538acf51cbc/src/main/scala/inox/solvers/z3/NativeZ3Optimizer.scala#L21-L25

```scala
  private type Target = targetProgram.type

  private class Underlying(override val program: Target,
                           override val context: Context)
                          (using override val semantics: targetSemantics.type)
```

This clearly should not change the semantics, so either the fix or the original error triggers a compilation bug. I'm working to minimize it, but it does not seem simple.

I'm not sure if it's best to stay on 3.3.0 or accept this fix for now; opinions are welcome.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ git.useGitDescribe := true
 
 organization := "ch.epfl.lara"
 
-scalaVersion := "3.3.0"
+scalaVersion := "3.3.3"
 
 scalacOptions ++= Seq(
   "-deprecation",
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.7.3
+sbt.version=1.10.0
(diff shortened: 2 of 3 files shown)


vkuncak:
Why did CI not complete here?

samuelchassot:
Because he is not in the trusted user list on Inox. Let me see if I can add him.

samuelchassot:
Now it's running

PR #212, opened 2024-05-21
Add --cvc-simple-models=true|false option that passes --model-cores=simple to cvc solvers

samuelchassot:


samuelchassot:
Tests are not passing, I didn't intend to start the CI, and I cannot kill it.

samuelchassot:
In the test suite `SolverSuite.scala`, some queries are now `Unknown` instead of `Sat`. For example:
```
v$0 != false && v$1 == () && v$2 != 'a' && v$3 != 0 && v$4 != 0 && v$5 != 0 && v$6 != x0000000000000 && v$7 != 0 && v$8 != 0 && v$9 != x000000000000000000000000000000000 && v$10 != 0 && v$11 != "" && v$12 != T#0 && !v$13.contains(0) && v$14(0) != 0 && v$15(0) != 0 && v$16(0) != 0 && v$17 != (0, false, 0)
```

Given this, I am not sure it is a good idea to use this option, as it seems to reduce the performance sufficiently to be noticable.

WDYT?

vkuncak:
@samuelchassot should we use this option only in cvc5 and not also in cvc4 ?

samuelchassot:
I thought about using it only with cvc5, but it is apparently supported by CVC4, so I guess we can just apply it to both when it is turned on on Stainless.

vkuncak:
I find it a bit hard to remember different variants for boolean flags. Could we have --simple-models=true/false , so people only need to know the name of the option if they expect it to be boolean?

vkuncak:
Can you please record this smt file? I'd like to send it to cvc devs.

> In the test suite `SolverSuite.scala`, some queries are now `Unknown` instead of `Sat`. For example:
> 
> ```
> v$0 != false && v$1 == () && v$2 != 'a' && v$3 != 0 && v$4 != 0 && v$5 != 0 && v$6 != x0000000000000 && v$7 != 0 && v$8 != 0 && v$9 != x000000000000000000000000000000000 && v$10 != 0 && v$11 != "" && v$12 != T#0 && !v$13.contains(0) && v$14(0) != 0 && v$15(0) != 0 && v$16(0) != 0 && v$17 != (0, false, 0)
> ```

samuelchassot:
> I find it a bit hard to remember different variants for boolean flags. Could we have --simple-models=true/false , so people only need to know the name of the option if they expect it to be boolean?

Sure!

samuelchassot:
> Can you please record this smt file? I'd like to send it to cvc devs.

I didn't manage yet to do it, I'll try again!

samuelchassot:
> I find it a bit hard to remember different variants for boolean flags. Could we have --simple-models=true/false , so people only need to know the name of the option if they expect it to be boolean?

Actually this is already the case, the usage is `--cvc-simple-models=true|false`. Good catch for the documentation I forgot to update though!

samuelchassot:
> I find it a bit hard to remember different variants for boolean flags. Could we have --simple-models=true/false , so people only need to know the name of the option if they expect it to be boolean?

This is already the case: stainless option is `--cvc-simple-models=true|false`. What you pointed at is the option that is passed to CVC. I can update the documentation is that's not clear. It is supposed to tell the user that this option is passed to the CVC solver

vkuncak:
> This is already the case: stainless option is `--cvc-simple-models=true|false`. What you pointed at is the option that is 

If that appears in the `--help` then all is good.

samuelchassot:
:)
It does! It appears like this:
```
[  Info  ] --cvc-simple-models[=true|false]                 Use --model-cores=simple with cvc family solvers to produce simpler models when queries are sat (default: false)
```

samuelchassot:
Here are the query files for cvc5 for the query that is `unknown` with `--model-cores=simple`, but `sat` without the option.

[smt2-files.zip](https://github.com/epfl-lara/inox/files/15465472/smt2-files.zip)

PR #215, opened 2024-08-09
Add option to ignore models during solving

sankalpgambhir:
Adding option `--ignore-models`, and relevant support for it in the unrolling solver.

Description (through Stainless):

```console
--ignore-models[=true|false]          Do not request models from solvers. (default: false)
                                      Overrides and disables check-models and feeling-lucky.
```

As expected, models cannot be checked if we do not ask the solver for them in the first place. Since `feeling-lucky` also uses models during the proof-checking phase in unrolling, it is also explicitly bypassed. The same models are also used for guiding further unrolling by discovering paths to prioritize, and thus this guidance is also bypassed here while the option is turned on.

With the option on, counterexamples are replaced by `(Empty Model)`. Example on a bolts example changed to be broken:

```scala
[info] [Warning ] /tmp/bolts/algorithms/sorting/QuickSortListOfBigInt.scala:31:21:  => INVALID
[info]                  case Nil() => Nil[BigInt]()
[info]                                ^^^^^^^^^^^^^
[info] [Warning ] Found counter-example:
[info] [Warning ]   (Empty model)
```

Functionality should be complete now. I have noticed no change to a slight improvement on smaller examples from `bolts/algorithms`, and I am continuing to do a more comprehensive check.

I am trying to see what would be reasonable tests to have for this option. Will request merging after that.

Feel free to check and comment on this!

### DIFF
--- a/src/it/scala/inox/solvers/SolvingTestSuite.scala
+++ b/src/it/scala/inox/solvers/SolvingTestSuite.scala
@@ -15,6 +15,7 @@ trait SolvingTestSuite extends TestSuite {
   } yield Seq(
     optSelectedSolvers(Set(solverName)),
     optCheckModels(checkModels),
+    optIgnoreModels(false),
     optAssumeChecked(assumeChecked),
     unrolling.optFeelingLucky(feelingLucky),
     unrolling.optUnrollAssumptions(unrollAssumptions),
--- a/src/main/scala/inox/Main.scala
+++ b/src/main/scala/inox/Main.scala
@@ -95,6 +95,7 @@ trait MainHelpers {
     solvers.optAssumeChecked -> Description(Solvers, "Assume that all impure expression have been checked"),
     solvers.optNoSimplifications -> Description(Solvers, "Disable selector/quantifier simplifications in solvers"),
     solvers.optCheckModels -> Description(Solvers, "Double-check counter-examples with evaluator"),
+    solvers.optIgnoreModels -> Description(Solvers, s"Do not request models from solvers.\nOverrides and disables ${optCheckModels.name} and ${unrolling.optFeelingLucky.name}."),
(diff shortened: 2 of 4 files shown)


sankalpgambhir:
Note that with this, the problematic example in epfl-lara/stainless#1549 can be checked:

```scala
sbt:stainless-dotty> run --solvers=smt-cvc5 /tmp/Dispenser.scala --vc-cache=false --ignore-models
[info] running (fork) stainless.Main --ignore-models --solvers=smt-cvc5 /tmp/Dispenser.scala
[...] // trimmed
[info] [  Info  ] ╟┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄╢
[info] [  Info  ] ║ total: 44   valid: 44   (0 from cache, 1 trivial) invalid: 0    unknown: 0    time:   13.43                                                   ║
[info] [  Info  ] ╚═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════╝
``` 

Without `--ignore-models`, an exception is thrown for now, as is described in the issue.

sankalpgambhir:
(Rebased to match with main)

PR #227, opened 2024-12-04
Fail silently solvers by default

samuelchassot:


samuelchassot:
@mario-bucev after tracking down the unsupported feature exception with Viktor, we finally found out that we had an option to do exactly what we needed the whole time :) ^^

samuelchassot:
@vkuncak that's ready to merge 👍

mario-bucev:
Aaaaah it was that simple 🙈 good find!

vkuncak:
Did you test this with Stainless? If so, perhaps make a Stainless PR to bump the Inox version?

PR #230, opened 2025-02-04
Fix cross-version conflict errors in the "docs" subproject.

iswoqqe:
# Goals
This PR addresses two issues:
- The project cannot be successfully imported in Intellij IDEA with the Scala plugin.
- The sbt command `docs/mdoc` does not run successfully.

Both currently fail with the error:
```
[error] Modules were resolved with conflicting cross-version suffixes in ProjectRef(uri("file:/home/iswoqqe/git/inox/"), "docs"):
[error]    org.scala-lang.modules:scala-parser-combinators _3, _2.13
```

# Changes
- The 2.13 version of the parser combinators dependency is now excluded in the "docs" subproject.
	- This makes it possible to import the project, but `docs/mdoc` still fails with a few different errors, which is why the changes below also are applied.
- Scala version of the "docs" subproject is increased to 3.5.2 (same as root project).
- The mdoc plugin version is increased to 2.6.3 (latest version).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,8 @@ git.useGitDescribe := true
 
 organization := "ch.epfl.lara"
 
-scalaVersion := "3.5.2"
+val inoxScalaVersion = "3.5.2"
+scalaVersion := inoxScalaVersion
 
 scalacOptions ++= Seq(
   "-deprecation",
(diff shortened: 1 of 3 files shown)


sankalpgambhir:
Thanks for the PR! Could you mark the PR as not-draft, so the CI can also run? Additionally, have a look at the CLA when you can. 

This is probably good reason to test the docs project in the CI as well going forward.

PR #248, opened 2026-02-27
Handle short form of z3 `(_ map op)` in model parsing

### LINKED ISSUE #236: Missformed SMT source in ((_ map or) (store ((as const (Array Int Bool)) false) 0 true)
With `Z3 version 4.14.1 - 64 bit` on macOS with M4 Pro, one test case fails:

```
[info] -  15: SAT - Sorted.scala-0.tip solver=no-inc:smt-z3 *** FAILED *** (98 milliseconds)
[info]   inox.solvers.smtlib.MissformedSMTException: Missformed SMT source in ((_ map or) (store ((as const (Array Int Bool)) false) 0 true) (store ((as const (Array Int Bool)) false) 26285 true)):
[info] Unknown SMT term of class: class smtlib.trees.Terms$FunctionApplication:
[info] ((_ map or) (store ((as const (Array Int Bool)) false) 0 true) (store ((as const (Array Int Bool)) false) 26285 true))
[info]   at inox.solvers.smtlib.SMTLIBParser.fromSMT(SMTLIBParser.scala:276
```

mbovel:
z3 sometimes returns `(_ map or)` instead of the fully-qualified `(_ map (or (Bool Bool) Bool))` form, and similarly for `and` and `ite`. This caused a `MissformedSMTException` when parsing models.

Fixes https://github.com/epfl-lara/inox/issues/236

### DIFF
--- a/src/main/scala/inox/solvers/smtlib/Z3Target.scala
+++ b/src/main/scala/inox/solvers/smtlib/Z3Target.scala
@@ -235,16 +235,22 @@ trait Z3Target extends SMTLIBTarget with SMTLIBDebugger {
         val set = fromSMT(arr, st)
         IfExpr(fromSMT(elem, BooleanType()), SetAdd(set, fromSMT(key, base)), set)
 
+      // Matches (_ map or) in both long form (_ map (or (Bool Bool) Bool))
+      // and short form (_ map or).
       case (FunctionApplication(
         QualifiedIdentifier(SMTIdentifier(SSymbol("map"),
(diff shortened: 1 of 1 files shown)


mbovel:
CI failure: `NoClassDefFoundError: smtlib/theories/FixedSizeBitVectors$SMod$`: it's a class from scala-smtlib that's missing at runtime because the staging cache is stale/corrupted. Should be fixed by #249.

mbovel:
Ping @sankalpgambhir

sankalpgambhir:
LGTM, thanks! Sorry I forgot to merge before.

PR #249, opened 2026-02-27
Isolate sbt global base per workspace in CI

mbovel:
scala-smtlib is fetched via sbt's `RootProject(uri(...))`, which
clones and compiles it into the sbt global staging directory. On
self-hosted runners this directory is shared across runners and
persists across runs, causing stale .tasty/.class files and race
conditions between concurrent jobs. Setting `SBT_GLOBAL_BASE` to a
workspace-local path isolates each job.

Can be reverted if scala-smtlib is published to Maven.


mbovel:
Okay, now seems to work ✅

samuelchassot (approved):

PR #250, opened 2026-02-27
Disable fail-fast in CI matrix

mbovel:
Keep all matrix jobs running even if one fails, so we get results from all platforms and JDK versions.


mbovel:
CI failures should be fixed by #249.

samuelchassot (approved):

samuelchassot:
We can merge when CI passes!

PR #260, opened 2026-06-09
BitVector <-> Integer conversions

sankalpgambhir:
Implements `BVToInt` and `IntToBV` for Princess and SMTLIB based solvers for now.

Some decisions had to be made to work with the signed/unsigned variants of BVs as they exist in Inox. SMTLIB BV logics have sized bitvectors, but no differentiation for signedness in the type, only in some operations.

Operations reduce to `bv2int` and `nat2bv` for SMTLIB and `cast2SignedBV`/`cast2UnsignedBV` for Princess.

I've added some simple tests but a general review of the conversions would be appreciated.


sankalpgambhir:
Oops I thought filter solvers had the opposite polarity

samuelchassot:
I don't see what we can do apart from what you did for the sign of BV, so it looks good to me 👍

PR #263, opened 2026-07-08
Add rlimit option for smt z3 and smt cvc5

samuelchassot:
From my experience, `--z3-rlimit=40000000` corresponds to around 10sec timeout on my Macbook with M3 Pro.


samuelchassot:
TODO
- [x] for now, the summary shows `unknown` when the resource limit is reached, instead of timeout.

samuelchassot:
The PR on Inox must be merged before ⚠️ https://github.com/epfl-lara/inox/pull/263

